PR #1, opened 2020-05-11
Make it build

dcristoloveanu:
Make it build


src/logging_stacktrace.c
```diff
@@ -0,0 +1,154 @@
...
+    while ((state = InterlockedCompareExchange(&doSymInit, 1, 0)) != 2)
+    {
+        if (state == 0)
+        {
+            (void)SymInitialize(GetCurrentProcess(), NULL, TRUE);
+            (void)InterlockedExchange(&doSymInit, 2);
+        }
+    }
+
+    char* result = malloc(TRACE_MAX_STACK_AS_STRING_SIZE);
```
mattdurak, line 64: >malloc [](start = 19, length = 6)

Should this be `logging_malloc`? Or is `logging_malloc` unused?

dcristoloveanu: It is OK to be malloc I think. If not, we'll revisit it. I am very weary of making any changes until this is simply replaced in all the repos.

PR #3, opened 2020-05-14
Add LogWarning and LogVerbose and clean up some of the old unused things

dcristoloveanu:
Add LogWarning and LogVerbose and clean up some of the old unused things.
Note: there is a separate store for reworking the entire logging, but right now what I need is Warning and Verbose so we can split out logs in 4 categories.

### DIFF
--- a/inc/azure_c_logging/xlogging.h
+++ b/inc/azure_c_logging/xlogging.h
@@ -16,21 +16,8 @@
 #include "azure_c_logging/logging_stacktrace.h"
 #endif
 
-#if defined(ESP8266_RTOS)
-#include "c_types.h"
-#endif
-
-#if defined(ARDUINO_ARCH_ESP8266)
-#include "esp8266/azcpgmspace.h"
-#endif
-
 #ifdef __cplusplus
-/* Some compilers do not want to play by the standard, specifically ARM CC */
-#ifdef MBED_BUILD_TIMESTAMP
-#include <stdio.h>
-#else
 #include <cstdio>
-#endif
 #else
 #include <stdio.h>
 #endif /* __cplusplus */
@@ -39,16 +26,13 @@
 extern "C" {
 #endif /* __cplusplus */
 
-#ifdef TIZENRT
-#undef LOG_INFO
-#endif
+#define LOG_CATEGORY_VALUES \
+    AZ_LOG_ERROR, \
+    AZ_LOG_WARNING, \
+    AZ_LOG_INFO, \
+    AZ_LOG_VERBOSE
 
-typedef enum LOG_CATEGORY_TAG
-{
-    AZ_LOG_ERROR,
-    AZ_LOG_INFO,
-    AZ_LOG_TRACE
-} LOG_CATEGORY;
+MU_DEFINE_ENUM(LOG_CATEGORY, LOG_CATEGORY_VALUES)
 
 #if defined _MSC_VER
 #define FUNC_NAME __FUNCDNAME__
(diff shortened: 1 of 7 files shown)


inc/azure_c_logging/xlogging.h
```diff
@@ -68,39 +52,15 @@ typedef void(*LOGGER_LOG_GETLASTERROR)(const char* file, const char* func, int l
...
-
-#define LogError LogInfo
-#define LOG(log_category, log_options, FORMAT, ...)  { \
-        static const char flash_str[] ICACHE_RODATA_ATTR STORE_ATTR = (FORMAT); \
-        printf(flash_str, ##__VA_ARGS__); \
-        printf("\r\n"); \
-}
-
-#else /* NOT ESP8266_RTOS */
+#else
```
mattdurak, line 63: >else [](start = 1, length = 4)

nit: can we comment what these else pre processor lines are for? I'm having trouble matching them up (could just be the diff). At least this one and the one on line 144

src/etwlogger_driver.c
```diff
@@ -169,6 +169,23 @@ static void perform_EventWriteLogErrorEvent(const char* content, const char* fil
...
+static void perform_EventWriteLogWarningEvent(const char* content, const char* file, const SYSTEMTIME* t, const char* func, int line)
+{
+    (void)t;
+    TraceLoggingWrite(g_hMyComponentProvider,
+        "LogError",
+        TraceLoggingLevel(TRACE_LEVEL_WARNING),
+        TraceLoggingString(file, "file"),
+        TraceLoggingString(func, "func"),
+        TraceLoggingInt32(line, "line"),
+        TraceLoggingString(content, "content")
```
mattdurak, line 181: >TraceLoggingString(content, "content") [](start = 8, length = 38)

Can we make "content" the first item (here and everywhere)? It makes it easier to view in perfview if you need to collect logs locally (because that only shows the first field).

src/etwlogger_driver.c
```diff
@@ -260,16 +289,28 @@ void etwlogger_log(LOG_CATEGORY log_category, const char* file, const char* func
     {
         switch (log_category)
         {
-            case AZ_LOG_INFO:
+            case AZ_LOG_ERROR:
```
mattdurak, line 292: >AZ_LOG_ERROR [](start = 17, length = 12)

consider combining ERROR/WARNING and INFO/VERBOSE since this is just checking if there is a file/func/line param or not

src/etwlogger_driver.c
```diff
@@ -260,16 +289,28 @@ void etwlogger_log(LOG_CATEGORY log_category, const char* file, const char* func
     {
```
mattdurak, line 289: Could cut down a bunch of lines if we just set a `const char* text_to_log` to either `text` or `"INTERNAL LOGGING ERROR: failed in vprintf_alloc"`.

Then we don't need 2 switch statements

inc/azure_c_logging/xlogging.h
dcristoloveanu: Added

---
In reply to: [425318306](https://github.com/Azure/azure-c-logging/pull/3#discussion_r425318306) [](ancestors = 425318306)

src/etwlogger_driver.c
dcristoloveanu: Sure

---
In reply to: [425321988](https://github.com/Azure/azure-c-logging/pull/3#discussion_r425321988) [](ancestors = 425321988)

src/etwlogger_driver.c
dcristoloveanu: Done.

---
In reply to: [425324491](https://github.com/Azure/azure-c-logging/pull/3#discussion_r425324491) [](ancestors = 425324491)

PR #11, opened 2020-07-12
logging work to not depend on pal (doesn't call malloc)

anporumb:


inc/azure_c_logging/xlogging.h
```diff
@@ -82,43 +82,37 @@ typedef void(*LOGGER_LOG_GETLASTERROR)(const char* file, const char* func, int l
...
-            {                                                                                                                                                           \
-                logger_function(log_category, __FILE__, FUNC_NAME, __LINE__, log_options, format, __VA_ARGS__);                                                                       \
+            char stackAsString[STACK_MAX_CHARACTERS];                                                                                                                   \
+            getStackAsString(stackAsString, sizeof(STACK_MAX_CHARACTERS));                                                                                              \
+            size_t formatSize = strlen(format);                                                                                                                         \
+            if(formatSize+sizeof(STACK_PRINT_FORMAT)+1>FORMAT_MAX_CHARACTERS)                                                                                           \
```
mattdurak, line 100: > if(formatSize+sizeof(STACK_PRINT_FORMAT)+1>FORMAT_MAX_CHARACTERS)    [](start = 11, length = 69)

nit: missing all spaces

src/etwlogger_driver.c
```diff
@@ -277,50 +145,61 @@ static void perform_EventWriteLogVerboseEvent(const char* message, const char* f
     );
 
 #if CALL_CONSOLE_LOGGER
-    consolelogger_log(AZ_LOG_VERBOSE, NULL, NULL, 0, LOG_LINE, "%s", message);
+    consolelogger_log(AZ_LOG_VERBOSE, file, func, line, LOG_LINE, "%s", message);
```
mattdurak, line 148: >file [](start = 38, length = 4)

We don't do this for Info above (maybe we should?). I think you want it in both or neither

src/etwlogger_driver.c
```diff
@@ -277,50 +145,61 @@ static void perform_EventWriteLogVerboseEvent(const char* message, const char* f
...
 
     va_list args;
     va_start(args, format);
 
-    lastErrorAsString = lastErrorToString(lastError);
-    if (lastErrorAsString == NULL)
-    {
-        char* userMessage = vprintf_alloc(format, args);
-        if (userMessage == NULL)
+    char message[LOG_SIZE_REGULAR * (LOG_SIZE_REGULAR>=sizeof(vsnprintf_failure_message))]; /*this construct will generate a compile time error (array of size 0) when LOG_SIZE_REGULAR is not enough to hold even the failure message*/
```
mattdurak, line 165: >(LOG_SIZE_REGULAR>=sizeof(vsnprintf_failure_message) [](start = 36, length = 52)

WTB more explicit static_assert macro (exists as built-in in C11 I guess but would be nice to have)

src/etwlogger_driver.c
```diff
@@ -277,50 +145,61 @@ static void perform_EventWriteLogVerboseEvent(const char* message, const char* f
...
+    {
+        (void)memcpy(lastErrorAsString, FormatMessageA_failure_message, sizeof(FormatMessageA_failure_message));
+    }
     else
     {
-        char* userMessage = vprintf_alloc(format, args);
-        if (userMessage == NULL)
+        /*remove extraneous newlines from FormatMessageA's output*/
+        char* whereAreThey;
+        if ((whereAreThey = strchr(lastErrorAsString, '\r')) != NULL)
```
mattdurak, line 191: Just feeling like consolelogger.c/etwlogger_driver.c have a lot of common code, maybe a shared helper for "strip newlines" is one example of code that could be shared. /consider

src/etwlogger_driver.c
```diff
@@ -219,7 +87,7 @@ static void perform_EventWriteLogErrorEvent(const char* content, const char* fil
 static void perform_EventWriteLogWarningEvent(const char* content, const char* file, const char* func, int line)
 {
     TraceLoggingWrite(g_hMyComponentProvider,
-        "LogError",
+        "LogWarning",
```
mattdurak, line 90: >LogWarning [](start = 9, length = 10)

ha good catch

tests/logging_int/main.c
```diff
@@ -6,10 +6,9 @@
 /*not much in the way of testing, but "should not crash" */
 int main(void)
 {
-
-    char really_big_string[4096]; /*4096 (the size of this array) needs top be bigger than LOG_SIZE_REGULAR from etwlogger_driver.c*/
+    char really_big_string[LOG_SIZE_REGULAR/2]; /*if the size of this array is greater than LOG_SIZE_REGULAR than it will not get printed as all memory is statically allocated and cannot grow*/
```
mattdurak, line 9: >*if the size of this array is greater than LOG_SIZE_REGULAR than it will not get printed as all memory is statically allocated and cannot grow*/ [](start = 49, length = 144)

Shouldn't we have a test for that?

inc/azure_c_logging/xlogging.h
anporumb: Added all the spaces

---
In reply to: [453742065](https://github.com/Azure/azure-c-logging/pull/11#discussion_r453742065) [](ancestors = 453742065)

src/etwlogger_driver.c
anporumb: Yes, good catch! More info > less info. 

---
In reply to: [453749301](https://github.com/Azure/azure-c-logging/pull/11#discussion_r453749301) [](ancestors = 453749301)

src/etwlogger_driver.c
anporumb: Microsoft's C has documented _STATIC_ASSERT([https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/static-assert-macro?view=vs-2019](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/static-assert-macro?view=vs-2019)) which seems to do the same, without providing a string literal like C++ since C++17.

_STATIC_ASSERT comes from from <crtdbg.h>.

I would like to keep it as is, as this solution is more portable (doesn't involve a specific header) than _STATIC_ASSERT which is Microsoft's own implementation. Also is rather canon in  C world to have an array of size 0 when something goes wrong.

---
In reply to: [453751150](https://github.com/Azure/azure-c-logging/pull/11#discussion_r453751150) [](ancestors = 453751150)

src/etwlogger_driver.c
anporumb: Yeah I agree, I think re-organizing logging to be more structured a definitely a task, I would like to track it with a different Pull Request.

---
In reply to: [453752517](https://github.com/Azure/azure-c-logging/pull/11#discussion_r453752517) [](ancestors = 453752517)

src/etwlogger_driver.c
anporumb: :)

---
In reply to: [453753003](https://github.com/Azure/azure-c-logging/pull/11#discussion_r453753003) [](ancestors = 453753003)

tests/logging_int/main.c
anporumb: We absolutely should. Added a test where the big string canot be printed because it exceeds the size of the buffer.

Also added LogWarning. LogVerbose and LogCritical to this file.

---
In reply to: [453754805](https://github.com/Azure/azure-c-logging/pull/11#discussion_r453754805) [](ancestors = 453754805)

PR #16, opened 2020-10-05
2 preprocessors

anporumb:


CMakeLists.txt
```diff
@@ -86,8 +86,8 @@ if(${logerror_captures_stack_traces})
     set(LOGGING_STACKTRACE_C_FILE ./src/logging_stacktrace.c)
     set(LOGGING_STACKTRACE_H_FILE ./inc/azure_c_logging/logging_stacktrace.h)
 else()
-    set(LOGGING_STACKTRACE_C_FILE)
-    set(LOGGING_STACKTRACE_H_FILE)
+    set(LOGGING_STACKTRACE_C_FILE ./src/logging_stacktrace_dummy.c)
+    set(LOGGING_STACKTRACE_H_FILE ./inc/azure_c_logging/logging_stacktrace.h)
```
jebrando, line 90: Why do you have an if with the same 2 lines in both places.  Should you just have the add_definitions line in the if and the set(LOGGING_STACKTRACE_C_FILES ... lines outside of the if?

jebrando (approved):

PR #17, opened 2020-10-12
fix bug where we write 1 character past the end of the array

anporumb:


src/logging_stacktrace.c
```diff
@@ -127,7 +127,7 @@ void getStackAsString(char* destination, size_t size)
             }
         }
     }
-    destination[0] = '\0';
+    destination[-1] = '\0';
```
dcristoloveanu, line 130: >-1 [](start = 16, length = 2)

Can you add a task in the things that never get done to write unit tests for this?
Otherwise cheers!

dcristoloveanu (approved):
:shipit:

src/logging_stacktrace.c
anporumb: https://msazure.visualstudio.com/One/_workitems/edit/8559770

anporumb: every line of code hides a bug...

PR #18, opened 2020-10-14
fix 1 more bug(?) when stack walk fails for frankly stack overflow re…

anporumb:
…asons internally and 1 more true bug of ours

### DIFF
--- a/inc/azure_c_logging/xlogging.h
+++ b/inc/azure_c_logging/xlogging.h
@@ -87,6 +87,15 @@ typedef void(*LOGGER_LOG_GETLASTERROR)(const char* file, const char* func, int l
 
 #define LOG(log_category, log_options, format, ...) MU_C2(LOG_,log_category)(log_category, log_options, format, ##__VA_ARGS__)
 
+#ifdef _MSC_VER
+static __declspec(thread) char stackAsString[STACK_MAX_CHARACTERS];
+static __declspec(thread) char formatWithStack[FORMAT_MAX_CHARACTERS];
+#else
+/*for C11 compilers*/
+static _Thread_local char stackAsString[STACK_MAX_CHARACTERS];
+static _Thread_local char formatWithStack[FORMAT_MAX_CHARACTERS];
+#endif
+
 #define LOG_AZ_LOG_WITH_STACK(log_category, log_options, format, ...)                                                                                                   \
 {                                                                                                                                                                       \
     (void)(0 && printf(format, ##__VA_ARGS__));                                                                                                                         \
@@ -94,7 +103,6 @@ typedef void(*LOGGER_LOG_GETLASTERROR)(const char* file, const char* func, int l
         LOGGER_LOG logger_function = xlogging_get_log_function();                                                                                                       \
         if (logger_function != NULL)                                                                                                                                    \
         {                                                                                                                                                               \
-            char stackAsString[STACK_MAX_CHARACTERS];                                                                                                                   \
             getStackAsString(stackAsString, sizeof(stackAsString));                                                                                                     \
             size_t formatSize = strlen(format);                                                                                                                         \
             if (formatSize + sizeof(STACK_PRINT_FORMAT) + 1 > FORMAT_MAX_CHARACTERS)                                                                                    \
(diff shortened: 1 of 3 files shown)


inc/azure_c_logging/xlogging.h
```diff
@@ -87,14 +87,22 @@ typedef void(*LOGGER_LOG_GETLASTERROR)(const char* file, const char* func, int l
 
 #define LOG(log_category, log_options, format, ...) MU_C2(LOG_,log_category)(log_category, log_options, format, ##__VA_ARGS__)
 
+#ifdef _MSC_VER
+static __declspec(thread) char stackAsString[STACK_MAX_CHARACTERS];
+static __declspec(thread) char formatWithStack[FORMAT_MAX_CHARACTERS];
+#else
+/*for C11 compilers*/
+static _Thread_local char stackAsString[STACK_MAX_CHARACTERS];
+static _Thread_local char formatWithStack[FORMAT_MAX_CHARACTERS];
```
dcristoloveanu, line 96: Consider a 
#ifdef _MSC_VER
#define XLOGGING_THREAD_LOCAL __declspec(thread)
#elif /* check for C11 compiler */
#define XLOGGING_THREAD_LOCAL _Thread_local
#else
#error
#endif

And then having the variables only once:

static XLOGGING_THREAD_LOCAL char stackAsString[STACK_MAX_CHARACTERS];
static XLOGGING_THREAD_LOCAL char formatWithStack[FORMAT_MAX_CHARACTERS];

There are actually 2 suggestions in here:
- Have a #define with the "thread local" modifier
- Error out at compile time of the compiler is not C11

inc/azure_c_logging/xlogging.h
```diff
@@ -87,14 +87,22 @@ typedef void(*LOGGER_LOG_GETLASTERROR)(const char* file, const char* func, int l
 
 #define LOG(log_category, log_options, format, ...) MU_C2(LOG_,log_category)(log_category, log_options, format, ##__VA_ARGS__)
 
+#ifdef _MSC_VER
+static __declspec(thread) char stackAsString[STACK_MAX_CHARACTERS];
+static __declspec(thread) char formatWithStack[FORMAT_MAX_CHARACTERS];
+#else
+/*for C11 compilers*/
+static _Thread_local char stackAsString[STACK_MAX_CHARACTERS];
```
dcristoloveanu, line 95: >static [](start = 0, length = 6)

Will this have one stackAsString for each translation unit?
Seems like we might want to have the variable and function that uses it to barf the stack in the .c file.

src/logging_stacktrace.c
```diff
@@ -38,8 +38,15 @@ static size_t memcat(char* destination, size_t destinationSize, const char* sour
 static const char SymFromAddrFailed[] = "SymFromAddr failed\n";
 static const char snprintfFailed[] = "snprintf failed\n";
 
+#ifdef _MSC_VER
+static __declspec(thread) void* stack[TRACE_MAX_STACK_FRAMES];
```
dcristoloveanu, line 42: >__declspec(thread) [](start = 7, length = 18)

Suggest XLOGGING_THREAD_LOCAL per the other comment

src/logging_stacktrace.c
```diff
@@ -52,82 +59,86 @@ void getStackAsString(char* destination, size_t size)
...
     AcquireSRWLockExclusive(&lockOverSymCalls);
 
     uint16_t numberOfFrames = CaptureStackBackTrace(1, TRACE_MAX_STACK_FRAMES, stack, NULL);
-    HANDLE process = GetCurrentProcess();
-
-    SYMBOL_INFO_EXTENDED symbolExtended;
-    SYMBOL_INFO* symbol = &symbolExtended.symbol;
+    if (numberOfFrames == 0)
+    {
+        snprintf(destination, destinationSize, "!CaptureStackBackTrace returned 0 frames");
```
dcristoloveanu, line 70: >snprintf [](start = 8, length = 8)

(void)? Theoretically it could also fail I guess. Theoretically, 'cause there's not much that should fail.

dcristoloveanu (approved):
:shipit:

src/logging_stacktrace.c
anporumb: added (void)

---
In reply to: [504394773](https://github.com/Azure/azure-c-logging/pull/18#discussion_r504394773) [](ancestors = 504394773)

inc/azure_c_logging/xlogging.h
anporumb: a VERY good observation.

logging absolutely needs a "rework". I've placed the variables appropriately now. 

---
In reply to: [504394400](https://github.com/Azure/azure-c-logging/pull/18#discussion_r504394400) [](ancestors = 504394400)

inc/azure_c_logging/xlogging.h
anporumb: I have the define. And using it.

I am not sure why a separate check would be needed for the other compilers: they would either understand _Thread_local C keyword or barf at compilation time anyway. Right?

---
In reply to: [504393960](https://github.com/Azure/azure-c-logging/pull/18#discussion_r504393960) [](ancestors = 504393960)

PR #20, opened 2020-10-29
use c build tools

anporumb:


.gitmodules
```diff
@@ -1,6 +1,6 @@
-[submodule "deps/azure-c-build-tools"]
-	path = deps/azure-c-build-tools
-	url = https://github.com/Azure/azure-c-build-tools.git
 [submodule "deps/azure-macro-utils-c"]
```
dcristoloveanu, line 1: >azure-macro-utils-c [](start = 17, length = 19)

Since this is at the leaf, don't you want to do this one now also and then pull both of them across all repos?

dcristoloveanu (approved):
:shipit:

.gitmodules
anporumb: I was thinking about that, and then I realized that once I approach zrpc/ebs there are going to be lost of changes, and I got scared that I might botch some. So this is just renaming/propagating azure-c-build tools.


And now thinking about it... maybe I should just do it. "on a second thought attempt" because then there are going to HOARDES of updates. 


muyeh, stay tuned!

---
In reply to: [514489956](https://github.com/Azure/azure-c-logging/pull/20#discussion_r514489956) [](ancestors = 514489956)

PR #32, opened 2021-01-16
Fix includes

dcristoloveanu:
Fix includes


build/devops_gated.yml
```diff
@@ -198,12 +206,60 @@ jobs:
...
+    displayName: 'Setup VS Vars'
+    inputs:
+      filename: '"c:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\Tools\VsDevCmd.bat"'
+      modifyEnvironment: true
+
+  - task: CMake@1
+    displayName: 'Running cmake'
+    inputs:
+      workingDirectory: 'build_ninja'
+      cmakeArgs: '.. -GNinja -DCMAKE_C_INCLUDE_WHAT_YOU_USE:UNINITIALIZED=include-what-you-use;-Xiwyu;--mapping_file=d:\foo.imp;--driver-mode=cl -DCMAKE_CXX_INCLUDE_WHAT_YOU_USE:UNINITIALIZED=include-what-you-use;-Xiwyu;--mapping_file=d:\foo.imp;--driver-mode=cl -Drun_unittests:bool=ON -Drun_int_tests:bool=ON -Drun_perf_tests:bool=ON -Duse_cppunittest:bool=ON -Drun_traceability:bool=OFF'
```
mattdurak, line 248: >d:\foo.imp [](start = 117, length = 10)

looks like some kind of example...

mattdurak (approved):
:shipit:

CMakeLists.txt
```diff
@@ -142,6 +138,7 @@ endif()
 
 target_link_libraries(c_logging ${c_logging_target_libs})
 
+# run_traceability is an option from c build tools
```
mattdurak, line 141: >run_traceability is an option from c build tools [](start = 2, length = 48)

I have a PR in c-build-tools to always create the target instead of this fix

build/devops_gated.yml
dcristoloveanu: It's the real thing. Just that the ./foo.imp was wrong. I was actually using foo.imp for local tests, but now I made it be used from c-build-tools.

PR #92, opened 2022-07-28
RFC for logging v2

dcristoloveanu:
This PR is an RFC for a v2 version of the logging library.
The focus for the v2 version is enabling contextual logging, so that logging contexts can be defined and inherited in order to allow for logging code to be easier to write and more consistent.

This is not intended to be the complete and final implementation, but rather an RFC to flush out heftier comments.

Some basic performance comparison with v1 shows that the contextual logging is not that heavy:

ETW+console:
v1 PERF_TEST_LOG_CRITICAL (0) 3410000 logs in 5.003131 s (5003130 us), 681573.239734 logs/us
v2 PERF_TEST_LOG_CRITICAL (0) 3410000 logs in 5.003131 s (5003130 us), 681573.239734 logs/us
v2 with local context created once PERF_TEST_LOG_CRITICAL (0) 3130000 logs in 5.004188 s (5004188 us), 625476.063722 logs/us
v2 with dynamic context created once PERF_TEST_LOG_CRITICAL (0) 3120000 logs in 5.012561 s (5012560 us), 622436.365158 logs/us
v2 with local context created every time PERF_TEST_LOG_CRITICAL (0) 2690000 logs in 5.010611 s (5010611 us), 536860.620708 logs/us
v2 with dynamic context created every time PERF_TEST_LOG_CRITICAL (0) 2690000 logs in 5.010611 s (5010611 us), 536860.620708 logs/us

ETW only:
v1 PERF_TEST_LOG_CRITICAL (0) 38860000 logs in 5.000658 s (5000658 us), 7770977.028565 logs/us
v2 PERF_TEST_LOG_CRITICAL (0) 38860000 logs in 5.000658 s (5000658 us), 7770977.028565 logs/us
v2 with local context created once PERF_TEST_LOG_CRITICAL (0) 36940000 logs in 5.000964 s (5000963 us), 7386576.606688 logs/us
v2 with dynamic context created once PERF_TEST_LOG_CRITICAL (0) 37310000 logs in 5.000699 s (5000698 us), 7460957.704209 logs/us
v2 with local context created every time PERF_TEST_LOG_CRITICAL (0) 13930000 logs in 5.002816 s (5002815 us), 2784431.919310 logs/us
v2 with dynamic context created every time PERF_TEST_LOG_CRITICAL (0) 13930000 logs in 5.002816 s (5002815 us), 2784431.919310 logs/us


v2/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+    add_dependencies(c_logging_v2_traceability traceabilitytool)
+endif()
+
+add_subdirectory(tests)
+
+if(${run_reals_check})
+    add_reals_check_target()
+endif()
+
+compileTargetAsC99(c_logging_v2)
```
anporumb, line 40: >compileTargetAsC99 [](http://example.com/codeflow?start=0&length=18)

I believe this logging will ripple everywhere.

I think it is time we ditch these very old remnants of "CompileasC99/C11" and simply just jump to c17 by default to all targets everywhere. (CXXFLAGS/CFLAGS?)

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
+# `c-logging` requirements
+
+## Basic functionality
+
+`c-logging` shall expose APIs allowing the user to log events that can be used for observing the software activity, debugging, etc.
+
+The exposeed API shall allow creating contexts both dynamically and on the stack and using those contexts for logging.
```
anporumb, line 7: >on the stack [](http://example.com/codeflow?start=68&length=12)

oooh, the mystery intensifies....

anporumb, line 7: >dynamically [](http://example.com/codeflow?start=52&length=11)

will be interesting to observe how that is achieved.

v2/inc/c_logging/logger.h
```diff
@@ -0,0 +1,24 @@
...
+    LOG_LEVEL_ERROR, \
+    LOG_LEVEL_WARNING, \
+    LOG_LEVEL_INFO, \
+    LOG_LEVEL_VERBOSE
+
+MU_DEFINE_ENUM_WITHOUT_INVALID(LOG_LEVEL, LOG_LEVEL_VALUES);
+
+#define LOG_MAX_MESSAGE_LENGTH 4096 /*in bytes - a message is not expected to exceed this size in bytes, if it does, only LOG_MAX_MESSAGE_LENGTH characters are retained*/
+
+void logger_log(LOG_LEVEL log_level, LOG_CONTEXT_HANDLE log_context, const char* format, ...);
```
anporumb, line 22: >void [](http://example.com/codeflow?start=0&length=4)

missing C++ thingies

anporumb, line 22: >logger_log [](http://example.com/codeflow?start=5&length=10)

as with any lib that has a elipsys API, it should have another one that does the same thing taking a va_list.

anporumb, line 22: >LOG_CONTEXT_HANDLE [](http://example.com/codeflow?start=37&length=18)

Interesting how the handle is not the first argument

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+// macro that can be used to create a dynamically allocated context
+#define LOG_CONTEXT_CREATE(dest_log_context, parent_context, ...) \
+    ...
+
+// destroy a dynamically allocated context
+#define LOG_CONTEXT_DESTROY(log_context_handle) \
+    ...
+
+// macro that can be used to create a context on the stack
+#define LOCAL_LOG_CONTEXT_DEFINE(log_context, parent_context, ...) \
```
anporumb, line 23: >LOCAL_LOG_CONTEXT_DEFINE [](http://example.com/codeflow?start=8&length=24)

poetry: let's name this LOG_CONTEXT_LOCAL_DEFINE.

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+
+```c
+    // chained allocated context
+    LOG_CONTEXT_HANDLE context_1;
+    LOG_CONTEXT_CREATE(context_1, NULL, LOG_CONTEXT_FIELD("name", "%s", "haga"));
+    LOG_CONTEXT_HANDLE context_2;
+    LOG_CONTEXT_CREATE(context_2, context_1, LOG_CONTEXT_FIELD("last name", "%s", "uaga"), LOG_CONTEXT_FIELD("age", "%d", 42));
+
+    logger_log(LOG_LEVEL_VERBOSE, context_1, "log context 1 data");
+    logger_log(LOG_LEVEL_VERBOSE, context_2, "log context 2 data");
```
anporumb, line 40: >logger_log [](http://example.com/codeflow?start=4&length=10)

would be nice to have in these examples some fields, like "log context 2 data %d %s", 2, "mumu"

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+
+    LOG_CONTEXT_DESTROY(context_1);
+
+    // stack allocated context
+    LOCAL_LOG_CONTEXT_DEFINE(local_context, context_2, LOG_CONTEXT_FIELD("the knights that say", "%s", "Nee!"));
```
anporumb, line 45: >context_2 [](http://example.com/codeflow?start=44&length=9)

Is context 2 available after context_1 was destroyed?

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+
+    logger_log(LOG_LEVEL_VERBOSE, &local_context, "log with local context");
```
anporumb, line 47: >local_context [](http://example.com/codeflow?start=35&length=13)

So local contexts need to have their address taken?

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+
+- `INFO`
+
+- `VERBOSE`
+
+## Sinks
+
+Multiple sinks shall be supported. Setting up globally which sinks to be used shall be supported (so that they do not have to be specified every time a `c_logging_log` call is made).
+
+Defining the sinks to use shall be done by defining a NULL terminated array, where each entry shall be a pointer to a structure that represents the sink interface.
```
anporumb, line 70: >by defining a NULL terminated array [](http://example.com/codeflow?start=40&length=35)

Interesting, in which context is sizeof(log_sinks) not available... 

Will this be passed on the stack?

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+
+```c
+const LOG_SINK* log_sinks[] = { &console_log_sink, &etw_log_sink, NULL };
+```
+
+### ETW sink
+
+An `ETW` sink shall be supported so that events are emitted via TraceLogging.
+
+For the `ETW` sink there should be a posibility of specifying the provider to use.
```
anporumb, line 82: >provider [](http://example.com/codeflow?start=66&length=8)

 provider (GUID)

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+
+`format` shall be a `printf`-like format string.
+
+`...` shall be the variable number of arguments required by `format`.
+
+## Contextual logging
+
+For each log line the library shall support specifying a context to be added to the log line.
+
+The context shall allow defining pairs of properties that are added to the event when the event is emitted.
```
anporumb, line 104: >properties [](http://example.com/codeflow?start=42&length=10)

property-value?

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+
+```c
+    LOCAL_LOG_CONTEXT_DEFINE(log_context, NULL, LOG_CONTEXT_FIELD("property_name", "%s", MU_P_OR_NULL(prop_value)));
+```
+
+Bonus:
+
+An ideal implementation for the stack allocated contexts is also to not do any memory copy until actually needed (until they are used).
+
+That would allow wring code like below while not performing the memory copies until the logging is actially executed.
```
anporumb, line 126: >wring [](http://example.com/codeflow?start=17&length=5)

writing

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+
+Example:
+
+```c
+    LOCAL_LOG_CONTEXT_DEFINE(log_context, NULL, LOG_CONTEXT_FIELD("property_name", "%s", MU_P_OR_NULL(prop_value)));
+```
+
+Bonus:
+
+An ideal implementation for the stack allocated contexts is also to not do any memory copy until actually needed (until they are used).
```
anporumb, line 124: >ideal [](http://example.com/codeflow?start=3&length=5)

but aren't stack memory allocation basically free?

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+That would allow wring code like below while not performing the memory copies until the logging is actially executed.
+
+```c
+void a(LOG_CONTEXT_HANDLE log_context, const char* correlation_id)
+{
+    LOCAL_LOG_CONTEXT_DEFINE(local_context, NULL, LOG_CONTEXT_FIELD("correlation_id", "%s", correlation_id));
+
+    if (malloc(42) == NULL)
+    {
+        logger_log(LOG_LEVEL_ERROR, local_context, "malloc failed");
```
anporumb, line 135: >logger_log [](http://example.com/codeflow?start=8&length=10)

can a child log into a parent handle? In this example, instead of local_context, could it log to "log_context"?

v2/inc/c_logging/logger.h
anporumb, line 22: >LOG_CONTEXT_HANDLE [](http://example.com/codeflow?start=37&length=18)

Is NULL a valid handle here? Sort of "the global unnamed log context", if it needs any creation, it can be achieved with init_one

v2/src/logger.c
```diff
@@ -0,0 +1,35 @@
...
+    va_start(args, format);
+
+    char message[LOG_MAX_MESSAGE_LENGTH];
+
+    vsprintf(message, format, args);
+
+    i = 0;
+    while (log_sinks[i] != NULL)
+    {
+        log_sinks[i++]->log_sink_log(log_level, log_context, message, __FILE__, __FUNCTION__, __LINE__);
```
anporumb, line 31: >__FUNCTION__ [](http://example.com/codeflow?start=80&length=12)

isn't __FUNCTION__ going to alway ebe "logger_log"? same for __LINE__, it will always be "31"

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,81 @@
...
+    (0xDAD29F36, 0x0A48, 0x4DEF, 0x9D, 0x50, 0x8E, 0xF9, 0x03, 0x6B, 0x92, 0xB4));
+/*DAD29F36-0A48-4DEF-9D50-8EF9036B92B4*/
+
+static volatile LONG isETWLoggerRegistered = 0;
+
+static void lazyRegisterEventProvider(void)
+{
+    /*lazily init the logger*/
+    LONG state;
+    while ((state = InterlockedCompareExchange(&isETWLoggerRegistered, 1, 0)) != 2) /*0 - not init, 1 - initializing, 2 - initialized*/
```
anporumb, line 29: >InterlockedCompareExchange [](http://example.com/codeflow?start=20&length=26)

should use lazy_init, i guess. Is it eve nuseable from logging? hmmm...

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,81 @@
...
+        event_name, \
+        TraceLoggingLevel(trace_level), \
+        TraceLoggingString(message, "content"), \
+        TraceLoggingString(file, "file"), \
+        TraceLoggingString(func, "func"), \
+        TraceLoggingInt32(line, "line")) \
+
+void log_sink_etw_log(LOG_LEVEL log_level, LOG_CONTEXT_HANDLE log_context, const char* message, const char* file, const char* func, int line)
+{
+    (void)log_context;
```
anporumb, line 60: >(void)log_context; [](http://example.com/codeflow?start=4&length=18)

hmmmmm :)

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+An ideal implementation for the stack allocated contexts is also to not do any memory copy until actually needed (until they are used).
+
+That would allow wring code like below while not performing the memory copies until the logging is actially executed.
+
+```c
+void a(LOG_CONTEXT_HANDLE log_context, const char* correlation_id)
+{
+    LOCAL_LOG_CONTEXT_DEFINE(local_context, NULL, LOG_CONTEXT_FIELD("correlation_id", "%s", correlation_id));
+
+    if (malloc(42) == NULL)
```
anporumb, line 133: >malloc [](http://example.com/codeflow?start=8&length=6)

I fear this might end up with some infinite recursion if logger_log calls malloc internally (and that fails).

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+Bonus:
+
+An ideal implementation for the stack allocated contexts is also to not do any memory copy until actually needed (until they are used).
+
+That would allow wring code like below while not performing the memory copies until the logging is actially executed.
+
+```c
+void a(LOG_CONTEXT_HANDLE log_context, const char* correlation_id)
+{
+    LOCAL_LOG_CONTEXT_DEFINE(local_context, NULL, LOG_CONTEXT_FIELD("correlation_id", "%s", correlation_id));
```
anporumb, line 131: >correlation_id [](http://example.com/codeflow?start=92&length=14)

when is "correlation_id" evaluated? At context creation time or at every logger_log?

One would "suppose" it is at every logger-log, otherwise, it being static, can just be a string (unless we care about data structure in the string).

But on the other hand, it is rather impossible to "store" things from one macro to the other, so... then it is statically evaluated.

Are these fields going to be added to the tracelogging structure and ... have some dynamic tracelogging?

v2/devdoc/logging.md
```diff
@@ -0,0 +1,162 @@
...
+Example:
+
+```c
+    LOG_CONTEXT_CREATE(dynamically_allocated_log_context, NULL, LOG_CONTEXT_FIELD("property_name", "%s", MU_P_OR_NULL(prop_value)));
+    LOG_CONTEXT_DESTROY(dynamically_allocated_log_context);
+```
+
+### Context chaining
+
+It shall be supported to chain contexts (define a context or create a context dynamically while specifying a parent context to inherit the information from).
```
anporumb, line 161: >chain [](http://example.com/codeflow?start=25&length=5)

Yeah, this is most awesome.

Would be really nice if we'd have in this document some spelled out example onf "what's on the screen" when logging in a chained context.

v2/src/logger.c
```diff
@@ -0,0 +1,35 @@
...
+
+void logger_log(LOG_LEVEL log_level, LOG_CONTEXT_HANDLE log_context, const char* format, ...)
+{
+    uint32_t i;
+    va_list args;
+    va_start(args, format);
+
+    char message[LOG_MAX_MESSAGE_LENGTH];
+
+    vsprintf(message, format, args);
```
anporumb, line 26: vsnprintf

v2/src/log_sink_console.c
```diff
@@ -0,0 +1,20 @@
...
+#include <stdio.h>
+
+#include "macro_utils/macro_utils.h"
+
+#include "c_logging/logger.h"
+#include "c_logging/log_context.h"
+
+#include "c_logging/log_sink_console.h"
+
+void log_sink_console_log(LOG_LEVEL log_level, LOG_CONTEXT_HANDLE log_context, const char* message, const char* file, const char* func, int line)
```
anporumb, line 13: >log_context [](http://example.com/codeflow?start=66&length=11)

seems to be allowed to be NULL (and that's good), but line 17 will dereference the NULL pointer.

v2/src/log_sink_console.c
```diff
@@ -0,0 +1,20 @@
...
+{
+    time_t t = time(NULL);
+
+    (void)printf("%s %s Time:%.24s File:%s Func:%s Line:%d %s\r\n", MU_ENUM_TO_STRING(LOG_LEVEL, log_level), (log_context != NULL) ? log_context->context_string : "", ctime(&t), file, func, line, message);
+}
+
+const LOG_SINK console_log_sink = { log_sink_console_log };
```
anporumb, line 20: >log_sink_console_log [](http://example.com/codeflow?start=36&length=20)

I've come to REALLY like lately designated initializers, so consider adding those :)

v2/src/log_sink_etw.c
anporumb: I was expecting that the context, being defined with "field" and "value", would be logged here also as collection of fields an dvalues. I suppose it would help in Geneva a lot.

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,81 @@
...
+#define ETW_TRACE_LOGGING_WRAPPER(event_name, trace_level, message, file, func, line) \
+    TraceLoggingWrite(g_my_component_provider, \
+        event_name, \
+        TraceLoggingLevel(trace_level), \
+        TraceLoggingString(message, "content"), \
+        TraceLoggingString(file, "file"), \
+        TraceLoggingString(func, "func"), \
+        TraceLoggingInt32(line, "line")) \
+
+void log_sink_etw_log(LOG_LEVEL log_level, LOG_CONTEXT_HANDLE log_context, const char* message, const char* file, const char* func, int line)
```
anporumb, line 58: >log_sink_etw_log [](http://example.com/codeflow?start=5&length=16)

looks like static

v2/tests/logger_int/main.c
```diff
@@ -0,0 +1,107 @@
...
+
+//#ifdef _MSC_VER /*linux doesn't have LogLastError*/
+//    /*LogLastError*/
+//    LogLastError("LogLastError: hello world!");
+//    LogLastError("LogLastError: a more complicated hello with int=%d.", 42);
+//    LogLastError("LogLastError: the most complicated hello with int=%d and a really big string that fits =%s.", 42, really_big_string_fits);
+//    LogLastError("LogLastError: the most complicated hello with int=%d and a really big string that does not fit=%s.", 42, really_big_string_doesnt_fit);
+//#endif
+    /*LogWarning*/
+    logger_log(LOG_LEVEL_WARNING, NULL, "LogWarning: hello world!");
```
anporumb, line 51: >LogWarning [](http://example.com/codeflow?start=41&length=10)

I think all these LogWarning should dissapear from here, as the level is given as argument to the function,

v2/tests/logger_int/main.c
```diff
@@ -0,0 +1,107 @@
...
+    logger_log(LOG_LEVEL_VERBOSE, NULL, "LogCritical: does PRI_BOOL work? %" PRI_BOOL "", MU_BOOL_VALUE(true) /*answer is yes, it works*/);
+    logger_log(LOG_LEVEL_VERBOSE, NULL, "LogCritical: does PRI_BOOL work with amazingly out of bounds value? %" PRI_BOOL "", MU_BOOL_VALUE(42));
+    logger_log(LOG_LEVEL_VERBOSE, NULL, "LogCritical: does PRI_BOOL works with false? %" PRI_BOOL "", MU_BOOL_VALUE(false));
+
+    char* prop_value = malloc(2);
+    (void)strcpy(prop_value, "a");
+
+    // allocate a context dynamically
+    LOG_CONTEXT_HANDLE log_context;
+    LOG_CONTEXT_CREATE(log_context, NULL, LOG_CONTEXT_FIELD("property_name", "%s", MU_P_OR_NULL(prop_value)));
```
anporumb, line 82: >LOG_CONTEXT_CREATE [](http://example.com/codeflow?start=4&length=18)

can this fail? asking for a friend.

v2/inc/c_logging/log_context.h
```diff
@@ -0,0 +1,105 @@
...
+#endif
+
+#include "macro_utils/macro_utils.h"
+
+// This code is intentionally a spike only to showcase an implementation
+// Code with tests shall be written after comments on the general idea are received
+
+typedef struct LOG_CONTEXT_TAG
+{
+    char* context_string;
```
anporumb, line 24: >context_string [](http://example.com/codeflow?start=10&length=14)

aha, so the context is an unstructured string.

v2/inc/c_logging/log_context.h
```diff
@@ -0,0 +1,105 @@
...
+
+static const char* log_context_get_string(LOG_CONTEXT_HANDLE log_context)
+{
+    return (log_context == NULL) ? NULL : log_context->context_string;
+}
+
+// macro that can be used to create a dynamically allocated context
+#define LOG_CONTEXT_CREATE(dest_log_context, parent_context, ...) \
+    uint32_t MU_C2(parent_context_string_length_, dest_log_context) = (uint32_t)log_context_get_string_length(parent_context); \
+    int MU_C2(chars_to_write_, dest_log_context); \
```
anporumb, line 71: >dest_log_context [](http://example.com/codeflow?start=31&length=16)

so this function has its name based on the *initial* context name.

How does one then pass contexts on the stack? That'd impose some serious global parameter names to be "with the same name".

Also, the following might not work?

```c
LOG_CONTEXT_HANDLE dynamic_log_context;
    LOG_CONTEXT_CREATE(dynamic_log_context, NULL, LOG_CONTEXT_FIELD("block_id", "%s", "pachoo"));

LOG_CONTEXT_HANDLE mumu=dynamic_log_context;

logger_log(LOG_LEVEL_CRITICAL, mumu, "LogCritical: hello world!");
```

?

v2/inc/c_logging/log_context.h
```diff
@@ -0,0 +1,105 @@
...
+    { \
+        int chars_written = 0; \
+        MU_FOR_EACH_1(GET_LOG_CONTEXT_FIELD_LENGTH, __VA_ARGS__) \
+        MU_C2(chars_to_write_, dest_log_context) = chars_written + MU_C2(parent_context_string_length_, dest_log_context) + 1; \
+    } \
+    dest_log_context = malloc(sizeof(LOG_CONTEXT) + MU_C2(chars_to_write_, dest_log_context) + 1); \
```
anporumb, line 77: >malloc [](http://example.com/codeflow?start=23&length=6)

malloc_2? or it doesn't exist in this context/\?

v2/inc/c_logging/log_context.h
```diff
@@ -0,0 +1,105 @@
...
+        dest_log_context->context_string = (void*)((LOG_CONTEXT*)(dest_log_context) + 1); \
+        { \
+            char* pos = dest_log_context->context_string + MU_C2(parent_context_string_length_, dest_log_context); \
+            const char* parent_context_string = log_context_get_string(parent_context); \
+            if (parent_context_string != NULL) \
+            { \
+                (void)memcpy(dest_log_context->context_string, parent_context_string, MU_C2(parent_context_string_length_, dest_log_context)); \
+            } \
+            int chars_written = 0; \
+            MU_FOR_EACH_1(STRINGIFY_LOG_CONTEXT_FIELD, __VA_ARGS__) \
```
anporumb, line 93: >1(STRINGIFY_LOG_CONTEXT_FIELD, __VA_ARGS__) \ [](http://example.com/codeflow?start=24&length=45)

this seems to produce a concatenation of "parent" and "child" context strings, how does one where the parent ended and the child began in the concatenated string? [also seeing that the context is an unstructured data]

v2/tests/logger_perf/main.c
```diff
@@ -0,0 +1,180 @@
...
+
+PERF_RESULT v1_results[MU_ENUM_VALUE_COUNT_WITHOUT_INVALID(PERF_TEST_LIST)];
+PERF_RESULT v2_results[MU_ENUM_VALUE_COUNT_WITHOUT_INVALID(PERF_TEST_LIST)];
+PERF_RESULT v2_results_local_context_once[MU_ENUM_VALUE_COUNT_WITHOUT_INVALID(PERF_TEST_LIST)];
+PERF_RESULT v2_results_dynamic_context_once[MU_ENUM_VALUE_COUNT_WITHOUT_INVALID(PERF_TEST_LIST)];
+PERF_RESULT v2_results_local_context_every_time[MU_ENUM_VALUE_COUNT_WITHOUT_INVALID(PERF_TEST_LIST)];
+PERF_RESULT v2_results_dynamic_context_every_time[MU_ENUM_VALUE_COUNT_WITHOUT_INVALID(PERF_TEST_LIST)];
+
+/*not much in the way of testing, but "should not crash" */
+int main(void)
```
anporumb, line 37: >main [](http://example.com/codeflow?start=4&length=4)

I think the measurements of these tests should be achieved with at least 1 session and 1 consumer connected to ehe ETW frameowrk. otherwise, ETW is very smart about dropping unwanted packets.

... also tests on the console /console + etw / etw only should have different numbers

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,81 @@
...
+    case LOG_LEVEL_WARNING:
+        ETW_TRACE_LOGGING_WRAPPER("LogWarning", TRACE_LEVEL_WARNING, message, file, func, line);
+    case LOG_LEVEL_INFO:
+        ETW_TRACE_LOGGING_WRAPPER("LogInfo", TRACE_LEVEL_INFORMATION, message, file, func, line);
+    case LOG_LEVEL_VERBOSE:
+        ETW_TRACE_LOGGING_WRAPPER("LogVerbose", TRACE_LEVEL_VERBOSE, message, file, func, line);
+    }
+}
+
+const LOG_SINK etw_log_sink = { log_sink_etw_log };
```
anporumb, line 81: >log_sink_etw_log [](http://example.com/codeflow?start=32&length=16)

will wait, here's reaons:
1) while this is an improvement over previous existing logging, i believe we should really think for 1 more day about having the context data:
a) structured
b) maybe evaluated every time when logged (so not only once at context creation time) - if at all possible
2) missing working examples with context being passed on the stack (and can a local context be passed as argument?)
3) try to avoid using malloc at all. Maybe have a define that says that no more than 10 contexts can be nested, and every "new" context is allocated on the stack and copies parent's chain of context.

... I am not sure I am helpful, but these were all my thoughts.

anporumb: aaah! and of course I am curious to see where this is going :)

anporumb (changes requested):
:clock1:

v2/src/log_sink_etw.c
anporumb: And one more thing: since I cannot add comments in the "descriotion.txt". Can you format those log/s in a more human readable format? I mean, if it does 600K/second, nobody cares about the 5th digit after the decimal point :)

v2/devdoc/logging.md
parth21999, line 126: >actially [](http://example.com/codeflow?start=99&length=8)

actually

v2/CMakeLists.txt
dcristoloveanu: Bumping to C17 is a separate orthogonal concern (unless we use something from it explicitly). Plus, this is also used in umock which in turn is used still by IoT SDK folks and others.
So, switching to C17 I really think should be a separate thing as I don't to take too much in the same change.

dcristoloveanu: ID Work Item Type Title State Assigned To Effort Remaining Work Original Estimate Value Area Iteration Path Tags
15085850 Product Backlog Item [Maintenance] Consider switching to C17 New         Business One\Custom\Azure Messaging Store\Iteartion 270

v2/inc/c_logging/logger.h
dcristoloveanu: Added

dcristoloveanu: A very good ideea, shall add, will leave the comment open until it's done.

dcristoloveanu: I think it won't matter which one is first, the log level or the context handle. So, I'll leave it as is for now.

v2/devdoc/logging.md
dcristoloveanu: Nothing that a rename cannot solve, done!

v2/devdoc/logging.md
dcristoloveanu: Added

v2/devdoc/logging.md
dcristoloveanu: Yes, it should be.

dcristoloveanu: Even if we do not copy we'd have to have some ref counting for it. I still think it is a good idea to have it available and not force some weird ownership rules on the user.

v2/devdoc/logging.md
dcristoloveanu: Yeah.

v2/devdoc/logging.md
dcristoloveanu: I was thinking at some point of allowing the configuration at link time.
As in having the user have a variable that is an array with the sinks. An the linker makes the connection - that's why the NULL terminated array.

For now it is anyway just an array of these sinks in a .c file in the logging lib. And I think we should start with such simple thing and then add if we need more.

v2/devdoc/logging.md
dcristoloveanu: Added

v2/devdoc/logging.md
dcristoloveanu: Changed.

v2/devdoc/logging.md
dcristoloveanu: Fixed.

v2/devdoc/logging.md
dcristoloveanu: Well, in order to make the contexts as const as we can we stringify the property-value pairs into a string that is held by the context.
That means copying memory, this is what I was referring to here.

Even if we do not stringify: there are 2 options: putting things in the context by value or by reference.

If we do by value memory copy happens also.
If we do by reference then there is no memory copy, but then the ownership will be iffy (user frees the memory => AV).

v2/inc/c_logging/log_context.h
dcristoloveanu: We don't have it in this context.

v2/src/log_sink_etw.c
dcristoloveanu: 1) a) I went back and forth about this a couple of times. I was thinking at some point to do a full blown type system like for ZRPC or umock. I think it would likely be overshoot.
The simplest thing is stringifying everything. Of course that is crappy from the point of view that not even the property-value pairs are separated.
So what I am thinking now we should at least separate the property-value pairs behind and make them all string. That would give us the ability to push the data somewhat more structured to Jarvis and get better queries.
Going full blown type system would likely not be that helpful because also ETW has a limited set of types and we'd end up anyway stringifying many things.

b) Let me tinker with this for a little bit. I am thinking if we would capture addresses for the values, that might also solve the desired deferred copy of memory due to stringifying.

dcristoloveanu: 2) I'll add that, should not be hard.

3) That's what the local context is for avoiding the malloc. I think chaining local contexts should be doable, let me get a working example of that.

dcristoloveanu: Very helpful btw. This is how most of our reviews should go. I feel like we lost some of the RFC part of our process that we used to have. Pride is a big enemy of writing SW.

dcristoloveanu: Also updated the perf test to not spew any digits. Btw the perf test will likely not get checked in in its current form, this was simply for me to prove to myself where it is and guide the development.

v2/src/log_sink_etw.c
dcristoloveanu: It's not available.
I am reluctant to pull more thingies in this repo, I'd rather build this with the least amount of dependencies (I thought at some point pulling in THANDLE :-))

v2/src/log_sink_etw.c
dcristoloveanu: Made static

v2/tests/logger_perf/main.c
dcristoloveanu: Agree, will do that on my next measurements

v2/devdoc/logging.md
dcristoloveanu: Fixed.

v2/tests/logger_int/main.c
dcristoloveanu: Removed indeed from here and the old int tests

v2/src/log_sink_console.c
dcristoloveanu: Done

v2/src/log_sink_console.c
dcristoloveanu: It's guarded by (log_context != NULL)

v2/inc/c_logging/logger.h
dcristoloveanu: Yes, NULL is allowed

v2/src/logger.c
dcristoloveanu: HAHA, I knew there was a reason I had at some point a macro as entry point! Now I remembered

dcristoloveanu: But me has an idea. Stay tuned

v2/devdoc/logging.md
dcristoloveanu: I'm not sure I got the idea. "a" is an example function (like our bsi_primary_internal_allocate_async.
And my point here was that it would be nice to have the context written once in this function and then used wherever a log error is needed for example.
And ideally it would be awesome if the work of stringifying and memory copy of the bytes if any are needed to only happen if the logging is actually executed.

v2/inc/c_logging/logger.h
dcristoloveanu: I will actually hold off on the va_args flavor for now.
That can always be added later (just so I get this moving).

v2/devdoc/logging.md
dcristoloveanu: Yes, that would be allowed (as long as log_context is in scope, it would work).

v2/src/logger.c
dcristoloveanu: Latest version has a macro for this to pick up the function, file and line no.

dcristoloveanu: I initially thought of modeling these as properties in a special context, but likely that'd be overthinking it.

v2/src/log_sink_etw.c
dcristoloveanu: Now it is.

v2/devdoc/logging.md
dcristoloveanu: It is evaluated at context definition/allocation time.
I think it would be somewhat misleading to evaluate it at logging time.

v2/inc/c_logging/log_context.h
dcristoloveanu: Right now there is no separation.
With the next version of what I did I believe we could add that.

v2/devdoc/logging.md
dcristoloveanu: Added output of example. in the .md

v2/src/logger.c
dcristoloveanu: Fixed.

v2/tests/logger_int/main.c
dcristoloveanu: The dynamic version can fail and the context is NULL in that case.

v2/inc/c_logging/log_context.h
dcristoloveanu: Now the fields in it have the concept of a type attached to them

v2/inc/c_logging/log_context.h
dcristoloveanu: This will need more work. But I think I might try it.

v2/inc/c_logging/log_context.h
dcristoloveanu: Contexts can be passed on the stack. copying contexts like mumu=dynamic_log_context; results in a problem of tracking the lifetime, since these contexts do not have ref counting.

Otherwise both cases described above work.

I might look into making these ref counted, but maybe we should not boil the ocean in one frankly.

v2/devdoc/logging.md
```diff
@@ -0,0 +1,221 @@
...
+// macro that can be used to create a dynamically allocated context
+#define LOG_CONTEXT_CREATE(dest_log_context, parent_context, ...) \
+    ...
+
+// destroy a dynamically allocated context
+#define LOG_CONTEXT_DESTROY(log_context_handle) \
+    ...
+
+// macro that can be used to create a context on the stack
+#define LOG_CONTEXT_LOCAL_DEFINE(log_context, parent_context, ...) \
```
anporumb, line 26: >LOG_CONTEXT_LOCAL_DEFINE [](http://example.com/codeflow?start=8&length=24)

Is this destroyed by the macro in line 22, or doesn't even have a destroy function? Even if it does not, should be spelled out explicitly.

v2/devdoc/logging.md
```diff
@@ -0,0 +1,221 @@
...
+    ...
+
+// when a user wants to specify properties for a context, they can use LOG_CONTEXT_PROPERTY
+// LOG_CONTEXT_PROPERTY(property_type, property_name, property_value)
```
anporumb, line 30: >LOG_CONTEXT_PROPERTY(property_type, property_name, property_value) [](http://example.com/codeflow?start=3&length=66)

One would assume that these properties are evaluated at their specification time, right? And sort-of copied... someplace until a log line is produced with these stamped properties of the context.

Can properties be replaced while SW is running?

Can the type of a property be replaced or is it defined only one?

If a property can be added to a context, how does one remove it from the context?

... I had questions.

v2/devdoc/logging.md
```diff
@@ -0,0 +1,221 @@
...
+
+// when a user wants to specify string fields in a context using printf-style formatting, they can use LOG_CONTEXT_STRING_PROPERTY
+// LOG_CONTEXT_STRING_PROPERTY(property_name, property_conversion_specifier, property_value)
```
anporumb, line 33: >property_conversion_specifier [](http://example.com/codeflow?start=46&length=29)

this is very interesting!!!! 

can I have the property_conversion_specifier as "%s=%d" and then pass "Popescu" and "42" as "property_value"?

Or is the field property_value limted to 1 value.

In that case... can be property conversion specifier be like "Address = %s"? (note the extra "Addres = ".

Or this is just a %s or %ls? If that's the case, we should have them separated, by LOG_CONTEXT_STRING_PROPERTY  /  LOG_CONTEXT_WSTRING_PROPERTY.

v2/devdoc/logging.md
```diff
@@ -0,0 +1,221 @@
...
+
+    LOGGER_LOG(LOG_LEVEL_VERBOSE, &local_context, "log with local context");
+
+    LOG_CONTEXT_DESTROY(context_2);
+```
+
+The above example produces:
+
+```
+LOG_LEVEL_CRITICAL Time:Thu Aug 11 12:51:55 2022 File:G:\w\c-logging\v2\tests\logger_int\main.c Func:main Line:96 name=haga last name=uaga age=42  log context 1 data, value of some_var = 42
```
anporumb, line 61: >last name=uag [](http://example.com/codeflow?start=124&length=13)

Why does "last name" appear here?

I think this line is produced by line 45 of this review which works exlcusively with context_1, so how does context_1 know about context_2 (where last name is defined in line 43 of this review)

v2/devdoc/logging.md
```diff
@@ -0,0 +1,221 @@
...
+Example usage:
+
+```c
+    // chained allocated context
+    LOG_CONTEXT_HANDLE context_1;
+    LOG_CONTEXT_CREATE(context_1, NULL, LOG_CONTEXT_STRING_PROPERTY(name, "%s", "haga"));
+    LOG_CONTEXT_HANDLE context_2;
+    LOG_CONTEXT_CREATE(context_2, context_1, LOG_CONTEXT_STRING_PROPERTY(last_name, "%s", "uaga"), LOG_CONTEXT_PROPERTY(int32_t, age, 42));
+
+    LOGGER_LOG(LOG_LEVEL_CRITICAL, context_1, "log context 1 data, value of some_var = %d", 42);
```
anporumb, line 45: >context [](http://example.com/codeflow?start=51&length=7)

calling these with "context" was in my mind unexpected, since this data (value of some var) is precisely not part of the context. 

Errr... I don't have a better poetry:(

Maybe just replace context with "handle"?

v2/devdoc/logging.md
```diff
@@ -0,0 +1,221 @@
...
+    LOGGER_LOG(LOG_LEVEL_VERBOSE, &local_context, "log with local context");
+
+    LOG_CONTEXT_DESTROY(context_2);
+```
+
+The above example produces:
+
+```
+LOG_LEVEL_CRITICAL Time:Thu Aug 11 12:51:55 2022 File:G:\w\c-logging\v2\tests\logger_int\main.c Func:main Line:96 name=haga last name=uaga age=42  log context 1 data, value of some_var = 42
+LOG_LEVEL_VERBOSE Time:Thu Aug 11 12:51:55 2022 File:G:\w\c-logging\v2\tests\logger_int\main.c Func:main Line:97 name=haga last name=uaga age=42  log context 2 data, some other string value is mumu
```
anporumb, line 62: >name=haga last name=uaga age=42 [](http://example.com/codeflow?start=113&length=31)

Would be nice if we could format (or chose how to format) these "context" property bags, maybe as JSON/XML or some other structured data, because at the moment they are completely unstructured (which is also fine). Depends on who reads them.

anporumb, line 62: > Line:97 [](http://example.com/codeflow?start=104&length=8)

Ooooooh, and while we're at it, can we please just combine Func and Line in one string? Like "main:97", would be awesome.

v2/devdoc/logging.md
```diff
@@ -0,0 +1,221 @@
...
+
+    LOGGER_LOG(LOG_LEVEL_CRITICAL, context_1, "log context 1 data, value of some_var = %d", 42);
+    LOGGER_LOG(LOG_LEVEL_VERBOSE, context_2, "log context 2 data, some other string value is %s", "mumu");
+
+    LOG_CONTEXT_DESTROY(context_1);
+
+    // stack allocated context
+    LOG_CONTEXT_LOCAL_DEFINE(local_context, context_2, LOG_CONTEXT_STRING_PROPERTY(the_knights_that_say, "%s", "Nee!"));
+
+    LOGGER_LOG(LOG_LEVEL_VERBOSE, &local_context, "log with local context");
```
anporumb, line 53: >local_context [](http://example.com/codeflow?start=35&length=13)

So with local context I always have to use &? :)

v2/devdoc/logging.md
```diff
@@ -0,0 +1,221 @@
...
+    LOG_CONTEXT_DESTROY(context_2);
+```
+
+The above example produces:
+
+```
+LOG_LEVEL_CRITICAL Time:Thu Aug 11 12:51:55 2022 File:G:\w\c-logging\v2\tests\logger_int\main.c Func:main Line:96 name=haga last name=uaga age=42  log context 1 data, value of some_var = 42
+LOG_LEVEL_VERBOSE Time:Thu Aug 11 12:51:55 2022 File:G:\w\c-logging\v2\tests\logger_int\main.c Func:main Line:97 name=haga last name=uaga age=42  log context 2 data, some other string value is mumu
+LOG_LEVEL_VERBOSE Time:Thu Aug 11 12:51:55 2022 File:G:\w\c-logging\v2\tests\logger_int\main.c Func:main Line:104 name=haga last name=uaga age=42 the_knights_that_say=Nee!  log with local context
+LOG_LEVEL_VERBOSE Time:Thu Aug 11 12:51:55 2022 File:G:\w\c-logging\v2\tests\logger_int\main.c Func:main Line:109 name=haga last name=uaga age=42 the_knights_that_say=Nee! other_knights_that_say=Moo!  log with nee and moo
```
anporumb, line 64: >log with nee and moo [](http://example.com/codeflow?start=201&length=20)

Where is this string coming from?

v2/devdoc/logging.md
```diff
@@ -0,0 +1,221 @@
...
+```
```
anporumb, line 65: I am starting to suspect that this output has not been updated with the latest code...

v2/devdoc/logging.md
```diff
@@ -0,0 +1,221 @@
...
+`LOG_CONTEXT_PROPERTY` is a macro that allows defining non-string properties by specifying a property type, name and a value.
+
+```c
+#define LOG_CONTEXT_PROPERTY(property_type, property_name, property_value) \
+    ...
+```
+
+The following types shall be supported out of the box:
+
+- ansi_charptr
```
anporumb, line 202: >ansi_charptr [](http://example.com/codeflow?start=2&length=12)

please consider having wchar_t too

v2/inc/c_logging/log_context.h
```diff
@@ -0,0 +1,193 @@
...
+
+typedef struct LOG_CONTEXT_TAG
+{
+    char* values_string;
+    uint32_t values_string_length;
+    uint8_t* values_data;
+    uint32_t values_data_length;
+    LOG_CONTEXT_PROPERTY_VALUE_PAIR* property_value_pairs_ptr;
+    uint32_t property_value_pair_count;
+    LOG_CONTEXT_PROPERTY_VALUE_PAIR property_value_pairs[];
```
anporumb, line 37: >property_value_pairs [](http://example.com/codeflow?start=36&length=20)

this and line 35 are different, or ... ?

v2/src/logger.c
```diff
@@ -0,0 +1,49 @@
...
+    va_start(args, format);
+
+    char message[LOG_MAX_MESSAGE_LENGTH];
+
+    int result = vsnprintf(message, LOG_MAX_MESSAGE_LENGTH, format, args);
+    if (result < 0)
+    {
+        (void)memcpy(message, logging_error_message, sizeof(logging_error_message));
+    }
+    else if (result == LOG_MAX_MESSAGE_LENGTH)
```
anporumb, line 32: >LOG_MAX_MESSAGE_LENGTH [](http://example.com/codeflow?start=23&length=22)

I am not sure that C standard says that vsnprintf will deploy any characters in the destitnation if it returns LOG_MAX_MESSAGE_LENGT.


Or I don't remember,...

v2/src/logger.c
```diff
@@ -0,0 +1,49 @@
...
+const LOG_SINK* log_sinks[] = { &console_log_sink, &etw_log_sink, NULL };
+const char logging_error_message[] = "Error generating log message";
+
+void logger_log(LOG_LEVEL log_level, LOG_CONTEXT_HANDLE log_context, const char* file, const char* func, int line_no, const char* format, ...)
+{
+    uint32_t i;
+    va_list args;
+    va_start(args, format);
+
+    char message[LOG_MAX_MESSAGE_LENGTH];
```
anporumb, line 25: there can be a static assert that verifies that the LOG_MESSAGE_SIZE can always accomodate logging_error_message.

v2/src/log_sink_console.c
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdio.h>
+#include <stdlib.h>
```
anporumb, line 5: >stdlib [](http://example.com/codeflow?start=10&length=6)

missing time.h

v2/src/log_sink_console.c
```diff
@@ -0,0 +1,30 @@
...
+
+#include "macro_utils/macro_utils.h"
+
+#include "c_logging/logger.h"
+#include "c_logging/log_context.h"
+
+#include "c_logging/log_sink_console.h"
+
+static void log_sink_console_log(LOG_LEVEL log_level, LOG_CONTEXT_HANDLE log_context, const char* message, const char* file, const char* func, int line)
```
anporumb, line 14: >log_sink_console_log [](http://example.com/codeflow?start=12&length=20)

would be awesome if critical/error would be "red" on the console (runs away)

v2/src/log_sink_console.c
```diff
@@ -0,0 +1,30 @@
...
+    time_t t = time(NULL);
+    char temp[LOG_MAX_MESSAGE_LENGTH];
+
+    size_t property_value_pair_count = log_context_get_property_value_pair_count(log_context);
+    LOG_CONTEXT_PROPERTY_VALUE_PAIR* property_value_pairs = (log_context != NULL) ? log_context_get_property_value_pairs(log_context) : NULL;
+    (void)printf("%s Time:%.24s File:%s Func:%s Line:%d ", MU_ENUM_TO_STRING(LOG_LEVEL, log_level), ctime(&t), file, func, line);
+    for (size_t i = 0; i < property_value_pair_count; i++)
+    {
+        property_value_pairs[i].type->to_string(&property_value_pairs[i], temp, LOG_MAX_MESSAGE_LENGTH);
+        (void)printf("%s=%s ", property_value_pairs[i].name, temp);
```
anporumb, line 25: >printf [](http://example.com/codeflow?start=14&length=6)

looks like printing these values from different threads leaves the console in a "haywire" condition. I believe there should be just 1 printf / log that contains everything.

v2/inc/c_logging/log_context.h
```diff
@@ -0,0 +1,193 @@
...
+    property_pair->type = &property_type ## _log_context_property_type; \
+    (void)chars_written;
+
+#define COPY_LOG_CONTEXT_FIELD(field_desc) \
+    MU_C2(EXPAND_, field_desc)
+
+// These macros have obviously grown to the point of being unwieldy
+// They clearly have to be rewritten and as much code as possible be moved out of being macro based
+// But for now they are here just for the spike purposes
+// I'd recommend not reviewing them a lot :-).
```
anporumb, line 72: >I'd recommend not reviewing them a lot :-). [](http://example.com/codeflow?start=3&length=43)

"shall skip" ;0

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,247 @@
...
+        for (uint32_t i = 0; i < property_value_count; i++)
+        {
+            size_t name_length = strlen(context_property_value_pairs[i].name);
+            (void)memcpy(pos, context_property_value_pairs[i].name, name_length + 1); pos += name_length + 1;
+            switch (context_property_value_pairs[i].type->get_type())
+            {
+            default:
+                *pos = TlgInANSISTRING;
+                break;
+            case LOG_CONTEXT_PROPERTY_TYPE_int32_t:
```
anporumb, line 129: >LOG_CONTEXT_PROPERTY_TYPE_int32_t [](http://example.com/codeflow?start=17&length=33)

likely needs more types here. I find all this code fascinating.

I cannot even imagine the amount of reverse engineering it took.

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,247 @@
...
+                EventDataDescCreate(&_tlgData[_tlgIdx], context_property_value_pairs[i].value, sizeof(INT64));
+                break;
+            case LOG_CONTEXT_PROPERTY_TYPE_ansi_charptr:
+                _tlgCreate1Sz_char(&_tlgData[_tlgIdx], context_property_value_pairs[i].value);
+                break;
+            }
+            _tlgIdx++;
+        }
+
+        // ... AND drumrolls, emit the event
```
anporumb, line 173: >AND drumrolls, emit the event [](http://example.com/codeflow?start=15&length=29)

:)))))))

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,247 @@
...
+    uint16_t _tlgEvtTag;
+    uint8_t metadata[];
+} SELF_DESCRIBED_EVENT;
+__pragma(pack(pop))
+
+#define MAX_SELF_DESCRIBED_EVENT_WITH_METADATA (sizeof(SELF_DESCRIBED_EVENT) + 4096)
+
+// This function is done with a little bit of reverse engineering of TraceLogging and guidance from 
+// the TraceLogging.h header about the format of the self described events
+static void emit_self_described_event(const char* event_name, uint32_t trace_level, LOG_CONTEXT_PROPERTY_VALUE_PAIR* context_property_value_pairs, uint32_t property_value_count, const char* message, const char* file, const char* func, int line)
```
anporumb, line 67: >context_property_value_pairs [](http://example.com/codeflow?start=117&length=28)

It looks like the structure of these properties is "flat", there's no hierarchy in the properties. Like "parent properties" /"grandparent /self.

Is that somethign tha tETW would support? Can ETW emit self-described structures?

It is awesome as is, and a great improvement1

v2/tests/logger_perf/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+#Copyright (c) Microsoft. All rights reserved.
+#Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+add_executable(logger_perf main.c timer.h timer_win32.c)
+target_link_libraries(logger_perf c_logging c_logging_v2)
+add_test(NAME logger_perf COMMAND logger_perf)
```
anporumb, line 6: >logger_perf [](http://example.com/codeflow?start=34&length=11)

I think for RFC level this is as good as it gets.

There's 2 minor outstanding issues:
1) logging examples seem to be old
2) whether there can be structured hierachically context (and therefore events)

But this is as good as it gets, seeing as the ETW has been reversed engineerd and disassembled :) 

Big cheers!

Will sign off.

anporumb (approved):
:shipit:

v2/inc/c_logging/log_context.h
dcristoloveanu: Yes, one is a variable size array to allocate everything in one shot for a context and the other in ln 35 is an alias.
In case of stack allocated contexts, the property_value_pairs_ptr alias actually points to a stack variable where the pairs are.

v2/devdoc/logging.md
dcristoloveanu: Added clarification

v2/devdoc/logging.md
dcristoloveanu: Yep. I could actually have it done as a pointer and have another variable that points to it. But for now it is likely good enough.

v2/src/log_sink_etw.c
dcristoloveanu: ETW can emit actually self described structures. It was also lurking in my mind how I could make use of that to group the properties of a given context. I don't think it should be that hard, so challenge accepted!

v2/src/log_sink_etw.c
dcristoloveanu: Absolutely, plan is to support almost all the ETW types (at least bool and GUID would be nice additions).
And it was quite fun. The kind of fun where you hit a brick wall that you can't do something and then you're like: REAAAALLY? I can't do what? Challenge accepted, after all it is all bytes! So I cannot deny that I really enjoyed it...

dcristoloveanu: Supporting all these will come as a separate PR (after the .md is merged in.

v2/devdoc/logging.md
dcristoloveanu: Updated example and output to reflect what was in the int test

v2/devdoc/logging.md
dcristoloveanu: rephrased a bit

v2/src/log_sink_console.c
dcristoloveanu: Added

v2/devdoc/logging.md
dcristoloveanu: Yes, was somewhat obsolete

v2/devdoc/logging.md
dcristoloveanu: Updated examples and output.

v2/src/log_sink_console.c
dcristoloveanu: This is SuCh A wOnDeRfUl IdEA!!!! Shall take it.

v2/devdoc/logging.md
dcristoloveanu: LOG_CONTEXT_STRING_PROPERTY(last name, "%s%s", "ua", "ga") has been added to the test.

dcristoloveanu: When we will have wchar_t_ptr as property type we'll also gain the LOG_CONTEXT_WSTRING_PROPERTY

v2/devdoc/logging.md
dcristoloveanu: Agree, but let's add that as a separate task.

dcristoloveanu: ID Work Item Type Title State Assigned To Effort Remaining Work Original Estimate Value Area Iteration Path Tags
15258058 Task Add wchar_t_ptr as type for v2 logging To Do           One\Custom\Azure Messaging Store\Iteration 271

v2/devdoc/logging.md
dcristoloveanu: Clarified that contexts are immutable once built.

dcristoloveanu: Maybe a future version will allow changing them, but let's start with this that the properties are placed in the contexts at construction time (be that dynamic or on the stack).

v2/devdoc/logging.md
dcristoloveanu: This sounds like a feature for the tasks that never get done.

dcristoloveanu: Let's consider this once we have the v2 (when time comes)

v2/src/log_sink_console.c
dcristoloveanu: Will do my best when implementing the console sink.
I guess we could always barf it in the temp that we anyway have here.

v2/src/log_sink_console.c
dcristoloveanu: Done.

v2/devdoc/logging.md
dcristoloveanu: Fixed in the console logger. I reckon the Geneva data we could leave with a separate field just in case we want to have any queries done on the separate line field. But if we feel we can change that at any point in the future too.

v2/src/logger.c
dcristoloveanu: Yeah this code will have tests checking the bounds.

v2/src/logger.c
dcristoloveanu: The vsnprintf function returns the number of characters that would have been written
had n been sufficiently large, not counting the terminating null character, oraneg ative
value if an encoding error occurred. Thus, the null-terminated output has been
completely written if and only if the returned value is nonnegative and less than n.

The way I read this if LOG_MAX_MESSAGE_LENGTH the null terminator is not written.

v2/inc/c_logging/log_context.h
dcristoloveanu: Structure has been added to console and etw

v2/src/log_sink_etw.c
dcristoloveanu: A POC of using structs in ETW has been done. perfview shows those decent, Geneva does not actually show the struct info :-). Maybe we can ask for some support for that at some point.

v2/tests/logger_perf/CMakeLists.txt
dcristoloveanu: Yay, i think I have addressed everything at RFC level.
Shall leave this PR for reference open until we are done with all the tasks, but will file PRs one by one for .md, headers, etc. separately.

jebrando:
It would be great to have an MD here that explains how to use this with some examples and a little explanation.

dcristoloveanu:
There actually is an .md, it is already in the repo.
So this RFC PR is diminishing as more and more things get merged. I am keeping the RFC PR as my stash until all is implemented.

---
In reply to: [1219008490](https://github.com/Azure/c-logging/pull/92#issuecomment-1219008490) [](http://example.com/codeflow?ancestors=1219008490,1219008490)

dcristoloveanu:
logging_v2 has been merged and is used.

PR #96, opened 2022-08-16
Add log sink header and spec

dcristoloveanu:


v2/devdoc/log_sink_requirements.md
```diff
@@ -0,0 +1,16 @@
+# `log_sink` requirements
+
+`log_sink` defines an interface for a log sink (as a structure that contains one function pointer).
+
+A log sink has the responsibility to output a logging event (be that to the console, producting an ETW event, writing it to a file, etc.)
+
+## Exposed API
+
+```
+typedef void (*LOG_SINK_LOG_FUNC)(LOG_LEVEL log_level, LOG_CONTEXT_HANDLE log_context, char* message, const char* file, const char* func, int line);
```
anporumb, line 10: >char [](http://example.com/codeflow?start=87&length=4)

why not const?

v2/inc/c_logging/log_sink.h
```diff
@@ -0,0 +1,16 @@
...
+#define LOG_SINK_H
+
+#include "c_logging/log_context.h"
+
+typedef void (*LOG_SINK_LOG_FUNC)(LOG_LEVEL log_level, LOG_CONTEXT_HANDLE log_context, char* message, const char* file, const char* func, int line);
+
+typedef struct LOG_SINK_TAG
+{
+    LOG_SINK_LOG_FUNC log_sink_log;
+} LOG_SINK;
```
anporumb, line 14: >LOG_SINK [](http://example.com/codeflow?start=2&length=8)

Is this a wanna-be interface? [with just 1 function?]

anporumb: line 3 of the .md says it is. Then consider naming it with `_IF`

anporumb (approved):
:shipit:

v2/devdoc/log_sink_requirements.md
dcristoloveanu: Made const

v2/inc/c_logging/log_sink.h
dcristoloveanu: Renamed to _IF. As part of that also extracted LOG_LEVEL in its own header.

PR #98, opened 2022-08-18
Add 32, 16 and 8 bit integers support

dcristoloveanu:
Add 32, 16 and 8 bit integers support

### DIFF
--- a/v2/devdoc/log_context_property_basic_types_requirements.md
+++ b/v2/devdoc/log_context_property_basic_types_requirements.md
@@ -82,4 +82,339 @@ typedef LOG_CONTEXT_PROPERTY_TYPE (*LOG_CONTEXT_PROPERTY_TYPE_GET_TYPE)(void);
 
 **SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_010: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int64_t).get_type` shall returns the property type `LOG_CONTEXT_PROPERTY_TYPE_int64_t`. **]**
 
-Note: the rest of the types will be implemented by cloning int32_t, thus will be done in a separate PR.
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string
+
+```c
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_TO_STRING)(void* property_value, char* buffer, size_t buffer_length);
+```
+
+`LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` produces the string representation of an `uint64_t`.
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_011: [** If `property_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` shall fail and return -1. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_012: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` shall call `snprintf` with `buffer`, `buffer_length` and format string `PRIu64` and pass in the values list the `uint64_t` value pointed to be `property_value`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_013: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` shall succeed and return the result of `snprintf`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_014: [** If any error is encountered, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` shall fail and return -1. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).copy
+
+```c
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_COPY)(void* dst_value, void* src_value);
+```
+
+`LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).copy` copies the `uint64_t` value from the address pointed by `src_value` to `dst_value`.
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_015: [** If `src_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_016: [** If `dst_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_017: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).copy` shall copy the bytes of the `uint64_t` value from the address pointed by `src_value` to `dst_value`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_018: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).copy` shall succeed and return 0. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).free
+
+```c
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_COPY)(void* value);
+```
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_019: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).free` shall return. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).get_type
+
+```c
+typedef LOG_CONTEXT_PROPERTY_TYPE (*LOG_CONTEXT_PROPERTY_TYPE_GET_TYPE)(void);
+```
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_020: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).get_type` shall returns the property type `LOG_CONTEXT_PROPERTY_TYPE_uint64_t`. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).to_string
+
+```c
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_TO_STRING)(void* property_value, char* buffer, size_t buffer_length);
+```
+
+`LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).to_string` produces the string representation of an `int32_t`.
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_021: [** If `property_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).to_string` shall fail and return -1. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_022: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).to_string` shall call `snprintf` with `buffer`, `buffer_length` and format string `PRId32` and pass in the values list the `int32_t` value pointed to be `property_value`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_023: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).to_string` shall succeed and return the result of `snprintf`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_024: [** If any error is encountered, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).to_string` shall fail and return -1. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).copy
+
+```c
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_COPY)(void* dst_value, void* src_value);
+```
+
+`LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).copy` copies the `int32_t` value from the address pointed by `src_value` to `dst_value`.
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_025: [** If `src_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_026: [** If `dst_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_027: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).copy` shall copy the bytes of the `int32_t` value from the address pointed by `src_value` to `dst_value`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_028: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).copy` shall succeed and return 0. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).free
+
+```c
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_COPY)(void* value);
+```
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_029: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).free` shall return. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).get_type
+
+```c
+typedef LOG_CONTEXT_PROPERTY_TYPE (*LOG_CONTEXT_PROPERTY_TYPE_GET_TYPE)(void);
+```
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_030: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).get_type` shall returns the property type `LOG_CONTEXT_PROPERTY_TYPE_int32_t`. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).to_string
+
+```c
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_TO_STRING)(void* property_value, char* buffer, size_t buffer_length);
+```
+
+`LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).to_string` produces the string representation of an `uint32_t`.
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_031: [** If `property_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).to_string` shall fail and return -1. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_032: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).to_string` shall call `snprintf` with `buffer`, `buffer_length` and format string `PRIu32` and pass in the values list the `uint32_t` value pointed to be `property_value`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_033: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).to_string` shall succeed and return the result of `snprintf`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_034: [** If any error is encountered, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).to_string` shall fail and return -1. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).copy
+
+```c
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_COPY)(void* dst_value, void* src_value);
+```
+
+`LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).copy` copies the `uint32_t` value from the address pointed by `src_value` to `dst_value`.
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_035: [** If `src_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_036: [** If `dst_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_037: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).copy` shall copy the bytes of the `uint32_t` value from the address pointed by `src_value` to `dst_value`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_038: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).copy` shall succeed and return 0. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).free
+
+```c
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_COPY)(void* value);
+```
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_039: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).free` shall return. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).get_type
+
+```c
+typedef LOG_CONTEXT_PROPERTY_TYPE (*LOG_CONTEXT_PROPERTY_TYPE_GET_TYPE)(void);
+```
+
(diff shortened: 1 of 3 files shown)


v2/devdoc/log_context_property_basic_types_requirements.md
```diff
@@ -82,4 +82,339 @@ typedef LOG_CONTEXT_PROPERTY_TYPE (*LOG_CONTEXT_PROPERTY_TYPE_GET_TYPE)(void);
...
+
+`LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` produces the string representation of an `uint64_t`.
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_011: [** If `property_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` shall fail and return -1. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_012: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` shall call `snprintf` with `buffer`, `buffer_length` and format string `PRIu64` and pass in the values list the `uint64_t` value pointed to be `property_value`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_013: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` shall succeed and return the result of `snprintf`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_014: [** If any error is encountered, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` shall fail and return -1. **]**
```
jebrando, line 99: >return -1 [](http://example.com/codeflow?start=148&length=9)

A pedantic Nitpick: Technically this is not entirely accurate.  The function snprintf will return a negative number if it fails it doesn't say anything in the spec or the Microsoft doc about what negative number it returns.  This would only matter if the user specifically is looking for -1 and not just a value that is negative.

jebrando (approved):
:shipit:

cyrus-jackson (approved):

v2/devdoc/log_context_property_basic_types_requirements.md
```diff
@@ -82,4 +82,339 @@ typedef LOG_CONTEXT_PROPERTY_TYPE (*LOG_CONTEXT_PROPERTY_TYPE_GET_TYPE)(void);
...
+
+```c
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_TO_STRING)(void* property_value, char* buffer, size_t buffer_length);
+```
+
+`LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` produces the string representation of an `uint64_t`.
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_011: [** If `property_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` shall fail and return -1. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_012: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string` shall call `snprintf` with `buffer`, `buffer_length` and format string `PRIu64` and pass in the values list the `uint64_t` value pointed to be `property_value`. **]**
```
anporumb, line 95: >snprintf [](http://example.com/codeflow?start=116&length=8)

One does have to wonder how much time does this logging take in our SW stack. And since we log pretty much everything, it miiiight be interesting to see if using something else except snprintf might be faster. 

Just thinking that snprintf can handle eeeeeeverything and we actually know beforehand the nature of the data to be stringified and the destination... maybe something else could be used for this purpose if performance is better with `stroull` for example, or even a hand crafted function.

v2/devdoc/log_context_property_basic_types_requirements.md
```diff
@@ -82,4 +82,339 @@ typedef LOG_CONTEXT_PROPERTY_TYPE (*LOG_CONTEXT_PROPERTY_TYPE_GET_TYPE)(void);
...
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_COPY)(void* dst_value, void* src_value);
+```
+
+`LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).copy` copies the `uint64_t` value from the address pointed by `src_value` to `dst_value`.
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_015: [** If `src_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_016: [** If `dst_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_017: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).copy` shall copy the bytes of the `uint64_t` value from the address pointed by `src_value` to `dst_value`. **]**
```
anporumb, line 113: >shall copy the bytes [](http://example.com/codeflow?start=99&length=20)

An interesting way to put it. Are we in a case where simple assignment cannot be used, because, let's say src/dst are unaligned and then we need memcpy? Is there a case where dst/src are overlapping and then we need memmove?

I would be very happy if src/dst are "somehow guaranteed by design" to be aligned and then compiler assignment just works.

v2/devdoc/log_context_property_basic_types_requirements.md
```diff
@@ -82,4 +82,339 @@ typedef LOG_CONTEXT_PROPERTY_TYPE (*LOG_CONTEXT_PROPERTY_TYPE_GET_TYPE)(void);
...
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_026: [** If `dst_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_027: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).copy` shall copy the bytes of the `int32_t` value from the address pointed by `src_value` to `dst_value`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_028: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).copy` shall succeed and return 0. **]**
+
+## LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).free
+
+```c
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_COPY)(void* value);
```
anporumb, line 168: >LOG_CONTEXT_PROPERTY_TYPE_COPY [](http://example.com/codeflow?start=14&length=30)

copy&Pasta detected.

anporumb: Seems like a case of here and everywhere

v2/devdoc/log_context_property_basic_types_requirements.md
```diff
@@ -82,4 +82,339 @@ typedef LOG_CONTEXT_PROPERTY_TYPE (*LOG_CONTEXT_PROPERTY_TYPE_GET_TYPE)(void);
...
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_COPY)(void* dst_value, void* src_value);
+```
+
+`LOG_CONTEXT_PROPERTY_TYPE_IMPL(int8_t).copy` copies the `int8_t` value from the address pointed by `src_value` to `dst_value`.
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_065: [** If `src_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int8_t).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_066: [** If `dst_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int8_t).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_067: [** `LOG_CONTEXT_PROPERTY_TYPE_IMPL(int8_t).copy` shall copy the bytes of the `int8_t` value from the address pointed by `src_value` to `dst_value`. **]**
```
anporumb, line 353: >bytes [](http://example.com/codeflow?start=112&length=5)

there's just 1 byte here :) a clear use for singular!

anporumb:
>     static int MU_C2(type_name,_log_context_property_to_string)(void* property_value, char* buffer, size_t buffer_size) \

size_t is SO annoying (because it is platform dependent)- can we have this as uint32_t? Just wondering. 

---
Refers to: v2/src/log_context_property_basic_types.c:15 in 03e601f. [](commit_id = 03e601f0b0bfe7db16545efa00476d34635a63f6, deletion_comment = False)

v2/src/log_context_property_basic_types.c
```diff
@@ -16,6 +16,13 @@
...
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_021: [ If property_value is NULL, LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).to_string shall fail and return -1. ]*/ \
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_031: [ If property_value is NULL, LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).to_string shall fail and return -1. ]*/ \
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_041: [ If property_value is NULL, LOG_CONTEXT_PROPERTY_TYPE_IMPL(int16_t).to_string shall fail and return -1. ]*/ \
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_051: [ If property_value is NULL, LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint16_t).to_string shall fail and return -1. ]*/ \
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_061: [ If property_value is NULL, LOG_CONTEXT_PROPERTY_TYPE_IMPL(int8_t).to_string shall fail and return -1. ]*/ \
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_071: [ If property_value is NULL, LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint8_t).to_string shall fail and return -1. ]*/ \
         if (property_value == NULL) \
```
anporumb, line 26: >property_value [](http://example.com/codeflow?start=12&length=14)

why is buffer=NULL a valid argument to the function?

v2/src/log_context_property_basic_types.c
```diff
@@ -27,6 +34,27 @@
...
+            /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_063: [ LOG_CONTEXT_PROPERTY_TYPE_IMPL(int8_t).to_string shall succeed and return the result of snprintf. ]*/ \
+            /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_064: [ If any error is encountered, LOG_CONTEXT_PROPERTY_TYPE_IMPL(int8_t).to_string shall fail and return -1. ]*/ \
+            /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_072: [ LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint8_t).to_string shall call snprintf with buffer, buffer_length and format string PRId8 and pass in the values list the uint8_t value pointed to be property_value. ]*/ \
+            /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_073: [ LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint8_t).to_string shall succeed and return the result of snprintf. ]*/ \
+            /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_074: [ If any error is encountered, LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint8_t).to_string shall fail and return -1. ]*/ \
             result = snprintf(buffer, buffer_size, "%" print_format_string "", *(type_name*)property_value); \
```
anporumb, line 58: >result = snprintf [](http://example.com/codeflow?start=12&length=17)

So this is interesting, as the caller of this `int` function will have to compare the returned result to see "did it fit" (returned result is less than "n") / "did it work" (returned result is greater than 0)

So this is really interesting, looking forward to the caller to see how the information is used!

v2/src/log_context_property_basic_types.c
```diff
@@ -50,8 +92,22 @@
...
+            /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_037: [ LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint32_t).copy shall copy the bytes of the uint64_t value from the address pointed by src_value to dst_value. ]*/ \
+            /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_047: [ LOG_CONTEXT_PROPERTY_TYPE_IMPL(int16_t).copy shall copy the bytes of the int64_t value from the address pointed by src_value to dst_value. ]*/ \
+            /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_057: [ LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint16_t).copy shall copy the bytes of the uint64_t value from the address pointed by src_value to dst_value. ]*/ \
+            /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_067: [ LOG_CONTEXT_PROPERTY_TYPE_IMPL(int8_t).copy shall copy the bytes of the int64_t value from the address pointed by src_value to dst_value. ]*/ \
+            /* Codes_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_077: [ LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint8_t).copy shall copy the bytes of the uint64_t value from the address pointed by src_value to dst_value. ]*/ \
             *(type_name*)dst_value = *(type_name*)src_value; \
```
anporumb, line 102: >*(type_name*)dst_value = *(type_name*)src_value; [](http://example.com/codeflow?start=12&length=48)

So this is the "shall copy the bytes". Consider renaming to "shall assign" :)

anporumb:
>         .to_string = MU_C2(type_name, _log_context_property_to_string), \

nice designated initializers!

---
Refers to: v2/src/log_context_property_basic_types.c:147 in 03e601f. [](commit_id = 03e601f0b0bfe7db16545efa00476d34635a63f6, deletion_comment = False)

anporumb:
>             result = vsnprintf(s, n, format, args);

ha ha ha ha

---
Refers to: v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c:68 in 03e601f. [](commit_id = 03e601f0b0bfe7db16545efa00476d34635a63f6, deletion_comment = False)

anporumb:
>     POOR_MANS_ASSERT(strcmp(buffer, "4") == 0);

So this verifies that the vsnprintf function works properly (and yes, both "4", and 2 are expected here). How's the user of the function going to look at 2?

1) I gave  a 2 byte buffer and more bytes were needed, so... I should do something about it

2) I have a 2 byte buffer and something is there... so I will take it as good and look the other way.

If "looking the other way" is an option, then the function doesn't even need to return the number of bytes written. Unless the number of bytes is used in advancing some buffer pointer... in which case... caller (the real one) would have to compare the passed "2" and the returned "2" and decide "there's been a truncation".

Just the random thought while reading this.

---
Refers to: v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c:175 in 03e601f. [](commit_id = 03e601f0b0bfe7db16545efa00476d34635a63f6, deletion_comment = False)

anporumb:
> }

around here I would have expected a test with int64_t_value=1 and buffer size = 2.

That is, the class of tests that says "when giving the exact minimum necessary number of bytes in the buffer, the function doesn't fail".

---
Refers to: v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c:177 in 03e601f. [](commit_id = 03e601f0b0bfe7db16545efa00476d34635a63f6, deletion_comment = False)

v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c
```diff
@@ -178,17 +178,17 @@ static void int32_t_to_string_with_truncation_succeeds(void)
 
 /* Tests_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_002: [ LOG_CONTEXT_PROPERTY_TYPE_IMPL(int64_t).to_string shall call snprintf with buffer, buffer_length and format string PRId64 and pass in the values list the int64_t value pointed to be property_value. ]*/
 /* Tests_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_003: [ LOG_CONTEXT_PROPERTY_TYPE_IMPL(int64_t).to_string shall succeed and return the result of snprintf. ]*/
-static void int32_t_to_string_with_negative_value_succeeds(void)
+static void int64_t_to_string_with_negative_value_succeeds(void)
```
anporumb, line 181: >with_negative [](http://example.com/codeflow?start=30&length=13)

NICE!!

v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c
```diff
@@ -178,17 +178,17 @@ static void int32_t_to_string_with_truncation_succeeds(void)
...
 
     setup_mocks();
     setup_expected_snprintf_call();
 
     // act
-    int result = LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).to_string(&int32_t_value, buffer, TEST_BUFFER_SIZE);
+    int result = LOG_CONTEXT_PROPERTY_TYPE_IMPL(int64_t).to_string(&int64_t_value, buffer, TEST_BUFFER_SIZE);
 
     // assert
     POOR_MANS_ASSERT(result > 0);
```
anporumb, line 194: >result > 0 [](http://example.com/codeflow?start=21&length=10)

There can be a better precision regarding the number of bytes, in this case it is "2".

v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c
```diff
@@ -272,35 +272,35 @@ static void int32_t_copy_succeeds(void)
...
 {
     // arrange
-    int32_t value = INT32_MAX;
+    int64_t value = INT64_MAX;
 
     setup_mocks();
 
     // act
-    LOG_CONTEXT_PROPERTY_TYPE_IMPL(int32_t).free(&value);
+    LOG_CONTEXT_PROPERTY_TYPE_IMPL(int64_t).free(&value);
```
anporumb, line 303: >free(&value) [](http://example.com/codeflow?start=44&length=12)

No linux nerdy compiler sees this as a warning? freeing the address of a stack variable... well... through some layers of function pointers... so maybe not.

v2/src/log_context_property_basic_types.c
anporumb: After reading some more code, it looks like this function wants to emulate snprintf totally, including NULL/0 arguments.

v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c
```diff
@@ -309,37 +309,1626 @@ static void int32_t_free_returns(void)
...
 
     // assert
-    POOR_MANS_ASSERT(result == LOG_CONTEXT_PROPERTY_TYPE_int32_t);
+    POOR_MANS_ASSERT(result == LOG_CONTEXT_PROPERTY_TYPE_int64_t);
     POOR_MANS_ASSERT(actual_and_expected_match);
 }
 
-/* very "poor man's" way of testing, as no test harness and mocking framework are available */
-int main(void)
+/* LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t_t).to_string */
```
anporumb, line 325: >uint64_t_t [](http://example.com/codeflow?start=34&length=10)

Is this some copy&pasta'd&replace'd type?

anporumb: An obvious case of here and everywhere

v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c
```diff
@@ -118,17 +118,17 @@ static void int32_t_to_string_with_NULL_value_fails(void)
 
 /* Tests_SRS_LOG_CONTEXT_PROPERTY_BASIC_TYPES_01_002: [ LOG_CONTEXT_PROPERTY_TYPE_IMPL(int64_t).to_string shall call snprintf with buffer, buffer_length and format string PRId64 and pass in the values list the int64_t value pointed to be property_value. ]*/
```
anporumb, line 119: >format string PRId64 [](http://example.com/codeflow?start=157&length=20)

with some more umock-y like data structures, this format string could be asserted too. Right now, it is not asserted.

v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c
```diff
@@ -309,37 +309,1626 @@ static void int32_t_free_returns(void)
...
+{
+    // arrange
+    char buffer[2];
+    uint64_t uint64_t_value = 42;
+
+    setup_mocks();
+    setup_expected_snprintf_call();
+
+    // act
+    int result = LOG_CONTEXT_PROPERTY_TYPE_IMPL(uint64_t).to_string(&uint64_t_value, buffer, 2);
```
anporumb, line 394: >2 [](http://example.com/codeflow?start=93&length=1)

this could be expressed as sizeof(buffer).

Consider here and in all the other tests.

anporumb (approved):
:shipit:

v2/devdoc/log_context_property_basic_types_requirements.md
dcristoloveanu: After I wrote a long dissertation I realized what you want to say.
I shall change this spec to say return a negative value.

v2/devdoc/log_context_property_basic_types_requirements.md
dcristoloveanu: Fixed.

v2/devdoc/log_context_property_basic_types_requirements.md
dcristoloveanu: Agree, but let's not optimize it ... yet. the ETW sink will actually not stringify these, so we'd optimize the console logs only. Let's create a task for it.

v2/devdoc/log_context_property_basic_types_requirements.md
dcristoloveanu: In the POC that I have they were not necessarily aligned, so, I guess I will leave it as is for now. If I get that context code to be aligned then we could come back to it.

v2/devdoc/log_context_property_basic_types_requirements.md
dcristoloveanu: Changed

v2/src/log_context_property_basic_types.c
dcristoloveanu: yep

v2/src/log_context_property_basic_types.c
dcristoloveanu: Oh, yes, I forgot I did the assign. I'll actually change this to memcpy.

v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c
dcristoloveanu: Fixed.

v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c
dcristoloveanu: The free for these basic types do nothing actually.
When we'd have a ref counted type for example (I secretly hope that we'll have custom types at some point), then the free for that one will have to do something.

v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c
dcristoloveanu: Yep, case of search/replace

v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c
dcristoloveanu: Added this validation

v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c
dcristoloveanu: Changed to sizeof(buffer) all over the place.

dcristoloveanu:
>     static int MU_C2(type_name,_log_context_property_to_string)(void* property_value, char* buffer, size_t buffer_size) \

I wanted to keep this close to the C functions that use size_t. Obviously at some point there will be a transition from uint32_t to size_t, I think I will leave that one in the caller.

---
In reply to: [1219722068](https://github.com/Azure/c-logging/pull/98#issuecomment-1219722068) [](http://example.com/codeflow?ancestors=1219722068,1219722068)

---
Refers to: v2/src/log_context_property_basic_types.c:15 in 03e601f. [](commit_id = 03e601f0b0bfe7db16545efa00476d34635a63f6, deletion_comment = False)

dcristoloveanu:
>             result = vsnprintf(s, n, format, args);

:-) How do you call that that does not exist :-)? Through another personality!

---
In reply to: [1219732247](https://github.com/Azure/c-logging/pull/98#issuecomment-1219732247) [](http://example.com/codeflow?ancestors=1219732247,1219732247)

---
Refers to: v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c:68 in 03e601f. [](commit_id = 03e601f0b0bfe7db16545efa00476d34635a63f6, deletion_comment = False)

dcristoloveanu:
>     POOR_MANS_ASSERT(strcmp(buffer, "4") == 0);

We'll need to use the result to advance.

---
In reply to: [1219736568](https://github.com/Azure/c-logging/pull/98#issuecomment-1219736568) [](http://example.com/codeflow?ancestors=1219736568,1219736568)

---
Refers to: v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c:175 in 03e601f. [](commit_id = 03e601f0b0bfe7db16545efa00476d34635a63f6, deletion_comment = False)

dcristoloveanu:
>         .to_string = MU_C2(type_name, _log_context_property_to_string), \

:-)

---
In reply to: [1219728105](https://github.com/Azure/c-logging/pull/98#issuecomment-1219728105) [](http://example.com/codeflow?ancestors=1219728105,1219728105)

---
Refers to: v2/src/log_context_property_basic_types.c:147 in 03e601f. [](commit_id = 03e601f0b0bfe7db16545efa00476d34635a63f6, deletion_comment = False)

dcristoloveanu:
> }

Added test

---
In reply to: [1219739437](https://github.com/Azure/c-logging/pull/98#issuecomment-1219739437) [](http://example.com/codeflow?ancestors=1219739437,1219739437)

---
Refers to: v2/tests/log_context_property_basic_types_ut/log_context_property_basic_types_ut.c:177 in 03e601f. [](commit_id = 03e601f0b0bfe7db16545efa00476d34635a63f6, deletion_comment = False)

PR #107, opened 2022-10-15
Add log sink console

dcristoloveanu:


v2/src/log_sink_console.c
```diff
@@ -1,18 +1,252 @@
...
+                /* Codes_SRS_LOG_SINK_CONSOLE_01_022: [ If any encoding error occurs during formatting of the line (i.e. if any printf class functions fails), log_sink_console.log_sink_log shall print Error formatting log line and return. ]*/
+                result = -1;
+                break;
+            }
+            else
+            {
+                snprintf_result = MIN(snprintf_result, (int)buffer_size);
+                buffer += snprintf_result;
+                buffer_size -= snprintf_result;
+                result += snprintf_result;
```
ACKPaul, line 70: Wouldn't it be nice to move these three lines of code to a function that returns  snprintf_result/buffer_size and use it everywhere?

v2/src/log_sink_console.c
```diff
@@ -1,18 +1,252 @@
...
+                    va_list args;
+                    va_start(args, message_format);
+                    int vsnprintf_result = vsnprintf(buffer, buffer_size, message_format, args);
+                    if (vsnprintf_result < 0)
+                    {
+                        /* Codes_SRS_LOG_SINK_CONSOLE_01_022: [ If any encoding error occurs during formatting of the line (i.e. if any printf class functions fails), log_sink_console.log_sink_log shall print Error formatting log line and return. ]*/
+                        error = true;
+                    }
+                    else
+                    {
```
ACKPaul, line 232: We can remove the else part .

ACKPaul (approved):

v2/src/log_sink_console.c
dcristoloveanu: Extracted in a macro.

v2/src/log_sink_console.c
dcristoloveanu: Added a comment to make it clear that we don't expect to do anything on that branch, but keeping the branch anyway to show it's not missed.

PR #123, opened 2023-05-10
Log etw sink with tests

dcristoloveanu:
Log etw sink with tests


v2/devdoc/log_sink_etw_requirements.md
```diff
@@ -20,160 +20,184 @@ The signature of `log_sink_etw.log_sink_log` is:
...
+**SRS_LOG_SINK_ETW_01_002: [** `log_sink_etw_log` shall maintain the state of whether `TraceLoggingRegister` was called in a variable accessed via `InterlockedXXX` APIs, which shall have 3 possible values: `NOT_REGISTERED` (1), `REGISTERING` (2), `REGISTERED`(3). **]**
 
-`log_sink_etw_log` shall use as provider GUID `DAD29F36-0A48-4DEF-9D50-8EF9036B92B4`.
+**SRS_LOG_SINK_ETW_01_003: [** `log_sink_etw_log` shall perform the below actions until the provider is registered or an error is encountered: **]**
 
-Note this can (and should) be improved to be configurable later.
+**SRS_LOG_SINK_ETW_01_004: [** If the state is `NOT_REGISTERED` (1): **]**
 
-`log_sink_etw_log` shall emit a self described event that shall have the name of the event as follows:
+- **SRS_LOG_SINK_ETW_01_005: [** `log_sink_etw_log` shall switch the state to `REGISTERING` (2). **]**
```
anporumb, line 33: >shall switch the state to `REGISTERING` (2 [](http://example.com/codeflow?start=52&length=42)

One does have to wonder "if we have platform_init"... why isn't this part registering part of platform_init.

... would save some "ifs" and some "interlocked" for every log that we do.

./consider for later.

v2/devdoc/log_sink_etw_requirements.md
```diff
@@ -20,160 +20,184 @@ The signature of `log_sink_etw.log_sink_log` is:
...
+**SRS_LOG_SINK_ETW_01_011: [** If the state is `REGISTERED` (3), `log_sink_etw_log` shall proceed to log the ETW event. **]**
 
-  - If `log_level` is `LOG_LEVEL_CRITICAL` the ETW logging level shall be `TRACE_LEVEL_CRITICAL`.
+**SRS_LOG_SINK_ETW_01_009: [** Checking and changing the variable that maintains whether `TraceLoggingRegister` was called shall be done using `InterlockedCompareExchange` and `InterlockedExchange`. **]**
 
-  - If `log_level` is `LOG_LEVEL_ERROR` the ETW logging level shall be `TRACE_LEVEL_ERROR`.
+**SRS_LOG_SINK_ETW_01_084: [** `log_sink_etw_log` shall use as provider GUID `DAD29F36-0A48-4DEF-9D50-8EF9036B92B4`. **]**
 
-  - If `log_level` is `LOG_LEVEL_WARNING` the ETW logging level shall be `TRACE_LEVEL_WARNING`.
+Note this can (and should) be improved to be configurable in a subsequent task.
```
anporumb, line 53: >task [](http://example.com/codeflow?start=74&length=4)

404 missing DevOps reference :)

v2/devdoc/log_sink_etw_requirements.md
```diff
@@ -20,160 +20,184 @@ The signature of `log_sink_etw.log_sink_log` is:
...
+- **SRS_LOG_SINK_ETW_01_048: [** If `log_context` is not `NULL`: **]**
 
-`log_sink_etw.log_sink_log` shall fill a `SELF_DESCRIBED_EVENT` structure, setting the following fields:
+  - **SRS_LOG_SINK_ETW_01_049: [** `log_sink_etw.log_sink_log` shall call `log_context_get_property_value_pair_count` to obtain the count of properties that are to be added to the ETW event. **]**
 
-- `_tlgBlobTyp` shall be set to `_TlgBlobEvent4`.
+  - **SRS_LOG_SINK_ETW_01_050: [** `log_sink_etw.log_sink_log` shall call `log_context_get_property_value_pairs` to obtain the properties that are to be added to the ETW event. **]**
 
-- `_tlgChannel` shall be set to 11.
+  - **SRS_LOG_SINK_ETW_01_082: [** If more than 64 properties are given in `log_context`, `log_sink_etw.log_sink_log` shall not add any properties to the event. **]**
```
anporumb, line 91: >shall not add any [](http://example.com/codeflow?start=118&length=17)

hm. 

Consider adding "some" is better than none at all?

v2/devdoc/log_sink_etw_requirements.md
```diff
@@ -20,160 +20,184 @@ The signature of `log_sink_etw.log_sink_log` is:
...
 
-- `_tlgLevel` shall be set to the appropriate logging level.
+**SRS_LOG_SINK_ETW_01_042: [** `log_sink_etw.log_sink_log` shall compute the metadata size for the self described event metadata as follows: **]**
 
-- `_tlgOpcode` shall be set to 0.
+- **SRS_LOG_SINK_ETW_01_043: [** Length of the event name (determined at compile time, excluding zero terminator) + 1.
+ **]**
```
anporumb, line 96: >  [](http://example.com/codeflow?start=0&length=1)

extraneous CRLF detected.

v2/devdoc/log_sink_etw_requirements.md
```diff
@@ -20,160 +20,184 @@ The signature of `log_sink_etw.log_sink_log` is:
...
+- **SRS_LOG_SINK_ETW_01_052: [** For struct properties one extra byte shall be added for the field count. **]**
 
-- The string `file` (as field name, excluding zero terminator), followed by one byte with the value `TlgInANSISTRING`.
+**SRS_LOG_SINK_ETW_01_085: [** If the size of the metadata and the formatted message exceeds 4096 bytes, `log_sink_etw.log_sink_log` shall not add any properties to the event. **]**
 
-- The string `func` (as field name, excluding zero terminator), followed by one byte with the value `TlgInANSISTRING`.
+**SRS_LOG_SINK_ETW_01_026: [** `log_sink_etw.log_sink_log` shall fill a `SELF_DESCRIBED_EVENT` structure, setting the following fields: **]**
 
-- The string `line` (as field name, excluding zero terminator), followed by one byte with the value `TlgInINT32`.
+- **SRS_LOG_SINK_ETW_01_027: [** `_tlgBlobTyp` shall be set to `_TlgBlobEvent4`. **]**
```
anporumb, line 113: >_tlgBlobTyp [](http://example.com/codeflow?start=34&length=11)

some reference about what this is and means would be nice.

v2/devdoc/log_sink_etw_requirements.md
```diff
@@ -20,160 +20,184 @@ The signature of `log_sink_etw.log_sink_log` is:
...
+  - **SRS_LOG_SINK_ETW_01_075: [** If the property type is `LOG_CONTEXT_PROPERTY_TYPE_uint16_t`, a byte with the value `TlgInUINT16` shall be added in the metadata. **]**
 
-For each property in `log_context`:
+  - **SRS_LOG_SINK_ETW_01_077: [** If the property type is `LOG_CONTEXT_PROPERTY_TYPE_int8_t`, a byte with the value `TlgInINT8` shall be added in the metadata. **]**
 
-- If the property type is `LOG_CONTEXT_PROPERTY_TYPE_int64_t`, the event data descriptor shall be filled with the value of the property by calling `EventDataDescCreate`.
+  - **SRS_LOG_SINK_ETW_01_079: [** If the property type is `LOG_CONTEXT_PROPERTY_TYPE_uint8_t`, a byte with the value `TlgInUINT8` shall be added in the metadata. **]**
 
-- If the property type is `LOG_CONTEXT_PROPERTY_TYPE_uint64_t`, the event data descriptor shall be filled with the value of the property by calling `EventDataDescCreate`.
+  - **SRS_LOG_SINK_ETW_01_056: [** If the property type is `LOG_CONTEXT_PROPERTY_TYPE_struct`, a byte with the value `_TlgInSTRUCT | _TlgInChain` shall be added in the metadata. **]**
```
anporumb, line 161: >LOG_CONTEXT_PROPERTY_TYPE_struct [](http://example.com/codeflow?start=60&length=32)

nice

anporumb:
> `log_sink_etw` implements a log sink interface that logs the events as ETW events.

Would be awesome if there were some pictures/text examples of the produced output.

---
Refers to: v2/devdoc/log_sink_etw_requirements.md:3 in 985d73d. [](commit_id = 985d73dffd4f4ebed1afc4f196f1366c7f3c3530, deletion_comment = False)

v2/devdoc/log_sink_etw_requirements.md
```diff
@@ -20,160 +20,184 @@ The signature of `log_sink_etw.log_sink_log` is:
...
+**SRS_LOG_SINK_ETW_01_040: [** `log_sink_etw.log_sink_log` shall set event data descriptor at index 2 by calling `_tlgCreate1Sz_char` with the value of the formatted message as obtained by using `printf` with the messages format `message_format` and the arguments in `...`. **]**
 
-- If the property type is `LOG_CONTEXT_PROPERTY_TYPE_uint8_t`, the event data descriptor shall be filled with the value of the property by calling `EventDataDescCreate`.
+**SRS_LOG_SINK_ETW_01_058: [** `log_sink_etw.log_sink_log` shall set event data descriptor at index 3 by calling `_tlgCreate1Sz_char` with `file`. **]**
 
-- If the property type is `LOG_CONTEXT_PROPERTY_TYPE_struct`, no event data descriptor shall be used.
+**SRS_LOG_SINK_ETW_01_059: [** `log_sink_etw.log_sink_log` shall set event data descriptor at index 4 by calling `_tlgCreate1Sz_char` with `func`. **]**
 
-- If the property type is any other value, it shall be ignored.
+**SRS_LOG_SINK_ETW_01_060: [** `log_sink_etw.log_sink_log` shall set event data descriptor at index 5 by calling `EventDataDescCreate` with `line`. **]**
```
anporumb, line 177: >EventDataDescCreate [](http://example.com/codeflow?start=114&length=19)

Why is this different than 01_059 or 01_058?

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+// max number of properties we want to have in the event
+#define LOG_MAX_ETW_PROPERTY_VALUE_PAIR_COUNT 64
+
+// max metadata size
+#define MAX_METADATA_SIZE 4096
+#define MAX_SELF_DESCRIBED_EVENT_WITH_METADATA (sizeof(SELF_DESCRIBED_EVENT) + MAX_METADATA_SIZE)
+
+/* Codes_SRS_LOG_SINK_ETW_01_084: [ log_sink_etw_log shall use as provider GUID DAD29F36-0A48-4DEF-9D50-8EF9036B92B4. ]*/
+TRACELOGGING_DEFINE_PROVIDER(
+    g_my_component_provider,
```
anporumb, line 28: >g_my_component_provider [](http://example.com/codeflow?start=4&length=23)

any chance we get a better name than "my"?

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+                }
+
+                /* Codes_SRS_LOG_SINK_ETW_01_008: [ log_sink_etw_log shall emit a LOG_LEVEL_INFO event as a self test , printing the fact that the provider was registered and from which executable (as obtained by calling _get_pgmptr). ]*/
+                log_sink_etw_log(LOG_LEVEL_INFO, NULL, __FILE__, __FUNCTION__, __LINE__, "ETW provider was registered succesfully (self test). Executable file full path name = %s", MU_P_OR_NULL(executable_full_path_name));
+
+                break;
+            }
+            else
+            {
+                (void)printf("ETW provider was NOT registered.\r\n");
```
anporumb, line 81: >registered [](http://example.com/codeflow?start=51&length=10)

could include the HRESULT as a hex number.

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+            {
+                /* Codes_SRS_LOG_SINK_ETW_01_007: [ log_sink_etw_log shall switch the state to REGISTERED. ]*/
+                (void)InterlockedExchange(&etw_provider_state, PROVIDER_STATE_REGISTERED);
+                /* Codes_SRS_LOG_SINK_ETW_01_008: [ log_sink_etw_log shall printf the fact that the provider was registered and from which executable (as obtained by calling _pgmptr). ]*/
+
+                char* executable_full_path_name;
+                if (_get_pgmptr(&executable_full_path_name) != 0)
+                {
+                    /* Codes_SRS_LOG_SINK_ETW_01_083: [ If _get_pgmptr fails, the executable shall be printed as UNKNOWN. ]*/
+                    executable_full_path_name = "UNKNOWN";
```
anporumb, line 71: >executable_full_path_name = "UNKNOWN"; [](http://example.com/codeflow?start=20&length=38)

How did this compile?

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+                }
+
+                /* Codes_SRS_LOG_SINK_ETW_01_008: [ log_sink_etw_log shall emit a LOG_LEVEL_INFO event as a self test , printing the fact that the provider was registered and from which executable (as obtained by calling _get_pgmptr). ]*/
+                log_sink_etw_log(LOG_LEVEL_INFO, NULL, __FILE__, __FUNCTION__, __LINE__, "ETW provider was registered succesfully (self test). Executable file full path name = %s", MU_P_OR_NULL(executable_full_path_name));
```
anporumb, line 75: >log_sink_etw_log [](http://example.com/codeflow?start=16&length=16)

so:
`internal_log_sink_etw_lazy_register_provider` calls `log_sink_etw_log` (here) and `log_sink_etw_log` calls `internal_log_sink_etw_lazy_register_provider` in line 436

Any chance of getting a stack overflow while the provider is still registering?

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+
+// want these aligned to 1 byte (copy/pasted from Tracelogging code)
+__pragma(pack(push, 1))
+typedef struct SELF_DESCRIBED_EVENT_TAG
+{
+    uint8_t _tlgBlobTyp;
+    uint8_t _tlgChannel;
+    uint8_t _tlgLevel;
+    uint8_t _tlgOpcode;
+    uint64_t _tlgKeyword;
```
anporumb, line 96: >_tlgKeyword [](http://example.com/codeflow?start=13&length=11)

Oh uh, who takes care of writing this?

1) the compiler does it because it is magic.
2) we do it byte by byte (cause I don't think this is aligned at this moment).

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+    uint16_t _tlgEvtMetaSize;
+    uint16_t _tlgEvtTag;
+    uint8_t metadata[];
+} SELF_DESCRIBED_EVENT;
+__pragma(pack(pop))
+
+// This function was written with a little bit of reverse engineering of TraceLogging and guidance from 
+// the TraceLogging.h header about the format of the self described events
+static void internal_emit_self_described_event(const char* event_name, uint16_t event_name_length, uint8_t trace_level, const LOG_CONTEXT_PROPERTY_VALUE_PAIR* context_property_value_pairs, uint32_t property_value_count, const char* message, const char* file, const char* func, int32_t line, va_list args)
```
anporumb, line 105: >uint32_t [](http://example.com/codeflow?start=189&length=8)

why not uint16_t?

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+        /* Codes_SRS_LOG_SINK_ETW_01_042: [ log_sink_etw.log_sink_log shall compute the metadata size for the self described event metadata as follows: ]*/
+        /* Codes_SRS_LOG_SINK_ETW_01_043: [ Length of the event name (determined at compile time, excluding zero terminator) + 1. ]*/
+        uint32_t i;
+
+        // alias to the event bytes
+        SELF_DESCRIBED_EVENT* self_described_event = (SELF_DESCRIBED_EVENT*)_tlgEvent;
+
+        uint8_t* pos = &self_described_event->metadata[0];
+
+        // at this point we filled all the event information, now we need to provide the metadata bytes
```
anporumb, line 125: >we filled all the event information [](http://example.com/codeflow?start=25&length=35)

I am a bit confused, as we don't seem to have filled anything yet, maybe later that happens.

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+    uint8_t metadata[];
+} SELF_DESCRIBED_EVENT;
+__pragma(pack(pop))
+
+// This function was written with a little bit of reverse engineering of TraceLogging and guidance from 
+// the TraceLogging.h header about the format of the self described events
+static void internal_emit_self_described_event(const char* event_name, uint16_t event_name_length, uint8_t trace_level, const LOG_CONTEXT_PROPERTY_VALUE_PAIR* context_property_value_pairs, uint32_t property_value_count, const char* message, const char* file, const char* func, int32_t line, va_list args)
+{
+    TraceLoggingHProvider const _tlgProv = g_my_component_provider;
+    if (trace_level < _tlgProv->LevelPlus1)
```
anporumb, line 108: >LevelPlus1 [](http://example.com/codeflow?start=32&length=10)

What makes this value change? :)

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+    {
+        // have a stack allocated buffer where we construct the event metadata and
+        uint8_t _tlgEvent[MAX_SELF_DESCRIBED_EVENT_WITH_METADATA];
```
anporumb, line 111: >MAX_SELF_DESCRIBED_EVENT_WITH_METADATA [](http://example.com/codeflow?start=26&length=38)

looks like we're always considering the `uint8_t metadata[];` field as `uint8_t metadata[4096];`.

Then why not stamp it directly in the struct? Would help with understanding.

anporumb: Also less casts.

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+        (void)memcpy(pos, event_name, event_name_length); pos += event_name_length;
+
+        // copy the field metadata (name and in type)
+
+        /* Codes_SRS_LOG_SINK_ETW_01_034: [ log_sink_etw.log_sink_log shall fill the event metadata: ]*/
+
+        /* Codes_SRS_LOG_SINK_ETW_01_035: [ The string content (as field name, excluding zero terminator), followed by one byte with the value TlgInANSISTRING. ]*/
+        /* Codes_SRS_LOG_SINK_ETW_01_044: [ Length of the content field name (determined at compile time, excluding zero terminator) + 1. ]*/
+        (void)memcpy(pos, "content", sizeof("content")); pos += sizeof("content");
+        *pos = TlgInANSISTRING;  pos++;
```
anporumb, line 137: >pos [](http://example.com/codeflow?start=33&length=3)

while it is pretty obvious that there's NO CHANCE for pos to escape after 4096 bytes, we still should have some static assert that verifies that metadata size can at least accomodate "content" + "file" + "func"+ ...

anporumb: It might be further helped by rearranging the fields so that "event_name" is the last one?

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+        (void)memcpy(pos, "file", sizeof("file")); pos += sizeof("file");
+        *pos = TlgInANSISTRING;  pos++;
+
+        /* Codes_SRS_LOG_SINK_ETW_01_037: [ The string func (as field name, excluding zero terminator), followed by one byte with the value TlgInANSISTRING. ]*/
+        /* Codes_SRS_LOG_SINK_ETW_01_046: [ Length of the func field name (determined at compile time, excluding zero terminator) + 1. ]*/
+        (void)memcpy(pos, "func", sizeof("func")); pos += sizeof("func");
+        *pos = TlgInANSISTRING;  pos++;
+
+        /* Codes_SRS_LOG_SINK_ETW_01_038: [ The string line (as field name, excluding zero terminator), followed by one byte with the value TlgInINT32. ]*/
+        /* Codes_SRS_LOG_SINK_ETW_01_047: [ Length of the line field name (determined at compile time, excluding zero terminator) + 1. ]*/
```
anporumb, line 150: >Length [](http://example.com/codeflow?start=44&length=6)

If you say "size of" then the +1 from the end of the spec is not needed.

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+
+        if (property_value_count <= LOG_MAX_ETW_PROPERTY_VALUE_PAIR_COUNT)
+        {
+            /* Codes_SRS_LOG_SINK_ETW_01_053: [ For each property in log_context the following shall be added to the event metadata: ]*/
+            for (i = 0; i < property_value_count; i++)
+            {
+                /* Codes_SRS_LOG_SINK_ETW_01_051: [ For each property in log_context, the length of the property name + 1 and one extra byte for the type of the field. ]*/
+                size_t name_length = strlen(context_property_value_pairs[i].name);
+                /* Codes_SRS_LOG_SINK_ETW_01_085: [ If the size of the metadata and the formatted message exceeds 4096 bytes, log_sink_etw.log_sink_log shall not add any properties to the event. ]*/
+                if ((uint16_t)(&self_described_event->metadata[MAX_METADATA_SIZE] - pos) < name_length + 1 + 1)
```
anporumb, line 164: > + 1 + 1 [](http://example.com/codeflow?start=102&length=8)

I suppose this can never overflow.

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+                {
+                    add_properties = false;
```
anporumb, line 166: >add_properties [](http://example.com/codeflow?start=20&length=14)

if we use printf for failure of registering the provider, maybe we could use it here too.

./here and everywhere

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+                    *pos = TlgInUINT8;
+                    break;
+
+                case LOG_CONTEXT_PROPERTY_TYPE_struct:
+                    /* Codes_SRS_LOG_SINK_ETW_01_052: [ For struct properties one extra byte shall be added for the field count. ]*/
+                    /* Codes_SRS_LOG_SINK_ETW_01_056: [ If the property type is LOG_CONTEXT_PROPERTY_TYPE_struct, a byte with the value _TlgInSTRUCT | _TlgInChain shall be added in the metadata. ]*/
+                    *pos = _TlgInSTRUCT | _TlgInChain;
+                    pos++;
+
+                    if (pos == &self_described_event->metadata[MAX_METADATA_SIZE])
```
anporumb, line 227: >self_described_event [](http://example.com/codeflow?start=32&length=20)

If line 172 executes and line 229 executes, doesn't that leave the serialized structure in an inconsistent state?

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+        else
+        {
+            metadata_size = (uint16_t)(pos - &self_described_event->metadata[0]);
+        }
+
+        char* formatted_message = (char*)pos;
+        size_t available_bytes = sizeof(_tlgEvent) - (formatted_message - (char*)self_described_event);
+
+        int formatted_message_length = vsnprintf(formatted_message, available_bytes, message, args);
+        if (formatted_message_length < 0)
```
anporumb, line 272: >formatted_message_length [](http://example.com/codeflow?start=12&length=24)

from C2x: `7.21.6.12` says `the null-terminated output has been completely written if and only if the
returned value is both nonnegative and less than n.`

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+        {
+            /* Codes_SRS_LOG_SINK_ETW_01_086: [ If any error occurs log_sink_etw.log_sink_log shall print Error emitting ETW event and return. ]*/
+            (void)printf("Error emitting ETW event.\r\n");
```
anporumb, line 275: >event [](http://example.com/codeflow?start=45&length=5)

maybe there's some enrichment we can have there as to add some strings or more information that would help tracking "which/what" event failed.

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+                if (add_properties)
+                {
+                    // did not fit, print again without properties and limit it
+                    /* Codes_SRS_LOG_SINK_ETW_01_085: [ If the size of the metadata and the formatted message exceeds 4096 bytes, log_sink_etw.log_sink_log shall not add any properties to the event. ]*/
+                    add_properties = false;
+                    metadata_size = metadata_size_without_properties;
+
+                    formatted_message = (char*)&self_described_event->metadata[metadata_size_without_properties];
+                    available_bytes = sizeof(_tlgEvent) - (formatted_message - (char*)self_described_event);
+                    formatted_message_length = vsnprintf(formatted_message, available_bytes, message, args);
```
anporumb, line 290: >vsnprintf [](http://example.com/codeflow?start=47&length=9)

So this is not OK here, as "args" seems to be consumed once in line 271 and once here.

Instead, a va_copy should be used

v2/devdoc/log_sink_etw_requirements.md
```diff
@@ -20,160 +20,184 @@ The signature of `log_sink_etw.log_sink_log` is:
...
+**SRS_LOG_SINK_ETW_01_085: [** If the size of the metadata and the formatted message exceeds 4096 bytes, `log_sink_etw.log_sink_log` shall not add any properties to the event. **]**
 
-- The string `func` (as field name, excluding zero terminator), followed by one byte with the value `TlgInANSISTRING`.
+**SRS_LOG_SINK_ETW_01_026: [** `log_sink_etw.log_sink_log` shall fill a `SELF_DESCRIBED_EVENT` structure, setting the following fields: **]**
 
-- The string `line` (as field name, excluding zero terminator), followed by one byte with the value `TlgInINT32`.
+- **SRS_LOG_SINK_ETW_01_027: [** `_tlgBlobTyp` shall be set to `_TlgBlobEvent4`. **]**
 
-For each property in `log_context` the following shall be added to the event metadata:
+- **SRS_LOG_SINK_ETW_01_028: [** `_tlgChannel` shall be set to 11. **]**
```
anporumb, line 115: general observation, while reading the code, maybe it belongs here, maybe in the code file, BUT!

Would be awesome to see some ASCII layout of the memory data structures we are building.

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+
+        int formatted_message_length = vsnprintf(formatted_message, available_bytes, message, args);
+        if (formatted_message_length < 0)
+        {
+            /* Codes_SRS_LOG_SINK_ETW_01_086: [ If any error occurs log_sink_etw.log_sink_log shall print Error emitting ETW event and return. ]*/
+            (void)printf("Error emitting ETW event.\r\n");
+        }
+        else
+        {
+            if ((size_t)formatted_message_length >= available_bytes)
```
anporumb, line 279: >if ((size_t)formatted_message_length >= available_bytes) [](http://example.com/codeflow?start=12&length=56)

very nice effort frankly.

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+                        // Ok as is
+                    }
+                }
+            }
+            else
+            {
+                // Ok as is
+            }
+
+            if (formatted_message_length >= 0)
```
anporumb, line 307: >>= [](http://example.com/codeflow?start=41&length=2)

Can this condition ever be "false"?

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+                self_described_event->_tlgEvtMetaSize = metadata_size + 4;
+
+                /* Codes_SRS_LOG_SINK_ETW_01_033: [ _tlgEvtTag shall be set to 128. ]*/
+                self_described_event->_tlgEvtTag = 128;
+
+                // now we need to fill in the event data descriptors
+                // first 2 are actually reserved for the event descriptor and metadata respectively
+                // all the rest starting at index 2 are actual data payloads in the event
+
+                /* Codes_SRS_LOG_SINK_ETW_01_039: [ log_sink_etw.log_sink_log shall fill an EVENT_DATA_DESCRIPTOR array of size 2 + 1 + 1 + 1 + 1 + property count. ]*/
```
anporumb, line 335: >1 [](http://example.com/codeflow?start=140&length=1)

I got lost at the third +1

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+                            /* Codes_SRS_LOG_SINK_ETW_01_062: [ If the property type is LOG_CONTEXT_PROPERTY_TYPE_struct, no event data descriptor shall be used. ]*/
+                            _tlgIdx--;
+                            break;
+                        }
+
+                        _tlgIdx++;
+                    }
+                }
+
+                // ... AND drumrolls, emit the event
```
anporumb, line 408: >drumrolls [](http://example.com/codeflow?start=27&length=9)

:))

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+
+        va_list args;
+        va_start(args, message_format);
+
+        switch (log_level)
+        {
+        default:
+            /* Codes_SRS_LOG_SINK_ETW_01_017: [ Otherwise the event name shall be Unknown. ]*/
+            /* Codes_SRS_LOG_SINK_ETW_01_024: [ Otherwise the ETW logging level shall be TRACE_LEVEL_NONE. ]*/
+            internal_emit_self_described_event(event_name_unknown, sizeof(event_name_unknown), TRACE_LEVEL_NONE, value_pairs, values_count, message_format, file, func, line, args);
```
anporumb, line 464: >values_count [](http://example.com/codeflow?start=126&length=12)

Last comment (almost):

Any chance the mighty "context" might provide some precomputed aggregated data (in general). So we don't have the serialize every time the same properties (get some "bulk" of them already in serialized form) with their numbers etc. Effectively context should do more in general, so not every `log_sink_etw_log` has to go through the same steps for the context.

v2/tests/log_sink_etw_2_ut/log_sink_etw_2_ut.c
```diff
@@ -0,0 +1,802 @@
...
+    MOCK_CALL_TYPE_TraceLoggingRegister_EventRegister_EventSetInformation, \
+    MOCK_CALL_TYPE__get_pgmptr, \
+    MOCK_CALL_TYPE__tlgCreate1Sz_char, \
+    MOCK_CALL_TYPE_EventDataDescCreate, \
+    MOCK_CALL_TYPE__tlgWriteTransfer_EventWriteTransfer, \
+    MOCK_CALL_TYPE_log_context_property_if_get_type
+
+MU_DEFINE_ENUM(MOCK_CALL_TYPE, MOCK_CALL_TYPE_VALUES)
+
+// very poor mans mocks :-(
```
anporumb, line 48: LOL

v2/tests/log_sink_etw_2_ut/log_sink_etw_2_ut.c
```diff
@@ -0,0 +1,802 @@
...
+        {
+            va_list args;
+            va_start(args, format);
+            // also capture the result in a variable for comparisons in tests
+            (void)vsnprintf(expected_calls[actual_call_count].u.printf_call.captured_output, sizeof(expected_calls[actual_call_count].u.printf_call.captured_output),
+                format, args);
+            va_end(args);
+
+            va_start(args, format);
+            result = vprintf(format, args);
```
anporumb, line 217: >vprintf [](http://example.com/codeflow?start=21&length=7)

Is thsi debug leftover?

anporumb: Actually doesn't look like that, it is more deep.

anporumb: If you don't want it on the screen, then you can write it in "NUL" file on windows (`/dev/nul` for Linux)`. Writes never fail there. Just saying.

anporumb: Maybe it is really intended to get on the screen and mimic the "real" calls. I got to that after seeing the next function,.

v2/tests/log_sink_etw_2_ut/log_sink_etw_2_ut.c
```diff
@@ -0,0 +1,802 @@
...
+
+    // assert
+    POOR_MANS_ASSERT(expected_call_count == actual_call_count);
+    POOR_MANS_ASSERT(actual_and_expected_match);
+}
+
+/* very "poor man's" way of testing, as no test harness and mocking framework are available */
+int main(void)
+{
+    when__get_pgmptr_fails_log_sink_etw_log_prints_executable_as_UNKNOWN();
```
anporumb, line 798: >when__get_pgmptr_fails_log_sink_etw_log_prints_executable_as_UNKNOWN [](http://example.com/codeflow?start=4&length=68)

So this whole file is intended to test "what happens when we cannot get the executable name"?

... crazy!

v2/tests/log_sink_etw_ut/log_sink_etw_mocked.c
```diff
@@ -0,0 +1,41 @@
...
+LONG mock_InterlockedCompareExchange(LONG volatile* Destination, LONG ExChange, LONG Comperand);
+LONG mock_InterlockedExchange(LONG volatile* Target, LONG Value);
+uint32_t mock_log_context_get_property_value_pair_count(LOG_CONTEXT_HANDLE log_context);
+const LOG_CONTEXT_PROPERTY_VALUE_PAIR* mock_log_context_get_property_value_pairs(LOG_CONTEXT_HANDLE log_context);
+TLG_STATUS mock_TraceLoggingRegister_EventRegister_EventSetInformation(const struct _tlgProvider_t* hProvider);
+errno_t mock__get_pgmptr(char** pValue);
+void mock__tlgCreate1Sz_char(PEVENT_DATA_DESCRIPTOR pDesc, char const* psz);
+void mock_EventDataDescCreate(PEVENT_DATA_DESCRIPTOR EventDataDescriptor, const VOID* DataPtr, ULONG DataSize);
+TLG_STATUS mock__tlgWriteTransfer_EventWriteTransfer(TraceLoggingHProvider hProvider, void const* pEventMetadata, LPCGUID pActivityId, LPCGUID pRelatedActivityId, UINT32 cData, EVENT_DATA_DESCRIPTOR* pData);
+int mock_vsnprintf(char* restrict s, size_t n, const char* restrict format, va_list arg);
```
anporumb, line 39: >restrict [](http://example.com/codeflow?start=25&length=8)

I see somebody went to C pdf and did copy&paste.

anporumb:
>     } u;

maybe this "u" can be omitted. It is still standard compliant I believe.

./consider

---
Refers to: v2/tests/log_sink_etw_ut/log_sink_etw_ut.c:159 in 985d73d. [](commit_id = 985d73dffd4f4ebed1afc4f196f1366c7f3c3530, deletion_comment = False)

anporumb:
>     } u;

here and everywhere if considered.

---
In reply to: [1544331142](https://github.com/Azure/c-logging/pull/123#issuecomment-1544331142) [](http://example.com/codeflow?ancestors=1544331142)

---
Refers to: v2/tests/log_sink_etw_ut/log_sink_etw_ut.c:159 in 985d73d. [](commit_id = 985d73dffd4f4ebed1afc4f196f1366c7f3c3530, deletion_comment = False)

anporumb:
>     expected_event_data_descriptors[2].Ptr = (ULONGLONG)"test";

Is this intended to be uintptr_t? Cause on x86... cough... might not be the same.

---
Refers to: v2/tests/log_sink_etw_ut/log_sink_etw_ut.c:3031 in 985d73d. [](commit_id = 985d73dffd4f4ebed1afc4f196f1366c7f3c3530, deletion_comment = False)

anporumb:
>     return asserts_failed;

We absolutely must have some INT tests.

I am not noticing any in this PR.

That and curiosity makes me want to look at it again at the next iteration.

---
Refers to: v2/tests/log_sink_etw_ut/log_sink_etw_ut.c:3094 in 985d73d. [](commit_id = 985d73dffd4f4ebed1afc4f196f1366c7f3c3530, deletion_comment = False)

anporumb (changes requested):
:clock1:

v2/devdoc/log_sink_etw_requirements.md
dcristoloveanu: Yeah that would be awesome.
ID Work Item Type Title State Assigned To Effort Remaining Work Original Estimate Value Area Iteration Path Tags
22152024 Task Initialize logging as part of platform_init to avoid interlocked calls To Do           One

dcristoloveanu: I had the same observation some time back for gballoc coming from some profile sessions.

v2/devdoc/log_sink_etw_requirements.md
dcristoloveanu: Added in the code.

v2/devdoc/log_sink_etw_requirements.md
```diff
@@ -20,160 +20,184 @@ The signature of `log_sink_etw.log_sink_log` is:
...
+- **SRS_LOG_SINK_ETW_01_047: [** Length of the `line` field name (determined at compile time, excluding zero terminator) + 1. **]**
 
-`log_sink_etw.log_sink_log` shall fill the event metadata:
+- **SRS_LOG_SINK_ETW_01_051: [** For each property in `log_context`, the length of the property name + 1 and one extra byte for the type of the field. **]**
 
-- The string `content` (as field name, excluding zero terminator), followed by one byte with the value `TlgInANSISTRING`.
+- **SRS_LOG_SINK_ETW_01_052: [** For struct properties one extra byte shall be added for the field count. **]**
 
-- The string `file` (as field name, excluding zero terminator), followed by one byte with the value `TlgInANSISTRING`.
+**SRS_LOG_SINK_ETW_01_085: [** If the size of the metadata and the formatted message exceeds 4096 bytes, `log_sink_etw.log_sink_log` shall not add any properties to the event. **]**
```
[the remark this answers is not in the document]
dcristoloveanu: Maybe, the problems are with ensuring we keep entire structures. Can be done of course, but would make things somewhat more complex.

Added a separate task for it.
ID Work Item Type Title State Assigned To Effort Remaining Work Original Estimate Value Area Iteration Path Tags
22152827 Task Consider keeping as many as possible of the properties in an event To Do Dan Cristoloveanu         One

v2/devdoc/log_sink_etw_requirements.md
dcristoloveanu: Considered and added a task

The problems are with ensuring we keep entire structures. Can be done of course, but would make things somewhat more complex.

Added a separate task for it.
ID Work Item Type Title State Assigned To Effort Remaining Work Original Estimate Value Area Iteration Path Tags
22152827 Task Consider keeping as many as possible of the properties in an event To Do Dan Cristoloveanu         One

v2/devdoc/log_sink_etw_requirements.md
dcristoloveanu: Fixed.

v2/devdoc/log_sink_etw_requirements.md
dcristoloveanu: Added

v2/devdoc/log_sink_etw_requirements.md
dcristoloveanu: Beacuse it is a diffferent type.

v2/src/log_sink_etw.c
dcristoloveanu: Renamed.

v2/src/log_sink_etw.c
dcristoloveanu: Done.

v2/src/log_sink_etw.c
dcristoloveanu: Compiles just fine actually.
And it is valid.

Example 8 on page 144 (section 6.7.9 Initialization) in C11 shows exactly this case.

However, it would be UB if this would get modified. So in order to avoid some confusion I introduced a const char*

v2/src/log_sink_etw.c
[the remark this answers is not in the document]
dcristoloveanu: It cannot, removed MU_P_OR_NULL

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+                            break;
+                        }
+
+                        _tlgIdx++;
+                    }
+                }
+
+                // ... AND drumrolls, emit the event
+                /* Codes_SRS_LOG_SINK_ETW_01_041: [ log_sink_etw.log_sink_log shall emit the event by calling _tlgWriteTransfer_EventWriteTransfer passing the provider, channel, number of event data descriptors and the data descriptor array. ]*/
+                _tlgWriteTransfer_EventWriteTransfer(_tlgProv, &self_described_event->_tlgChannel, ((void*)0), ((void*)0), _tlgIdx, _tlgData);
```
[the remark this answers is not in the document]
dcristoloveanu: NULL-ified

v2/tests/log_sink_etw_2_ut/log_sink_etw_2_ut.c
dcristoloveanu: I just wanted to have the "real" function called, yes.

v2/src/log_sink_etw.c
dcristoloveanu: va_copy deployed

v2/tests/log_sink_etw_ut/log_sink_etw_mocked.c
dcristoloveanu: Yes, they did!

dcristoloveanu:
>     return asserts_failed;

INT tests are planned with this task part of the same PBI.
Need to break them a bit.

[https://msazure.visualstudio.com/One/_backlogs/backlog/Azure%20Messaging%20Store/Backlog%20items/?workitem=15258065](https://msazure.visualstudio.com/One/_backlogs/backlog/Azure%20Messaging%20Store/Backlog%20items/?workitem=15258065)

---
In reply to: [1544336481](https://github.com/Azure/c-logging/pull/123#issuecomment-1544336481) [](http://example.com/codeflow?ancestors=1544336481,1544336481)

---
Refers to: v2/tests/log_sink_etw_ut/log_sink_etw_ut.c:3094 in 985d73d. [](commit_id = 985d73dffd4f4ebed1afc4f196f1366c7f3c3530, deletion_comment = False)

v2/src/log_sink_etw.c
dcristoloveanu: Made it uint16_t.

v2/src/log_sink_etw.c
dcristoloveanu: Yes, it can, if we attempt to printf the message in line 293 and it fails with a negative return then we can't do much more about it.

dcristoloveanu:
>     expected_event_data_descriptors[2].Ptr = (ULONGLONG)"test";

It does compile on x86 too. We have the leg that compiles it. I'll leave it as such as opposed to littering it with casts that could be removed ... When it borks then we can add the cast.

---
In reply to: [1544335198](https://github.com/Azure/c-logging/pull/123#issuecomment-1544335198) [](http://example.com/codeflow?ancestors=1544335198,1544335198)

---
Refers to: v2/tests/log_sink_etw_ut/log_sink_etw_ut.c:3031 in 985d73d. [](commit_id = 985d73dffd4f4ebed1afc4f196f1366c7f3c3530, deletion_comment = False)

dcristoloveanu:
>     } u;

Removed in this PR.
Also added a task to remove in all the others as I don't want to keep adding changes to this PR.

ID Work Item Type Title State Assigned To Effort Remaining Work Original Estimate Value Area Iteration Path Tags
22184349 Task Remove union names in unit test files for c-logging To Do           One  

---
In reply to: [1544331477](https://github.com/Azure/c-logging/pull/123#issuecomment-1544331477) [](http://example.com/codeflow?ancestors=1544331477,1544331142)

---
Refers to: v2/tests/log_sink_etw_ut/log_sink_etw_ut.c:159 in 985d73d. [](commit_id = 985d73dffd4f4ebed1afc4f196f1366c7f3c3530, deletion_comment = False)

v2/src/log_sink_etw.c
dcristoloveanu: Yes, got moved during some refactoring, so it was now stale, fixed it.

v2/src/log_sink_etw.c
dcristoloveanu: The guts of Tracelogging. I did check it takes effect by running the perf tests. And I plan some INT tests for it in the next PRs.

v2/tests/log_sink_etw_2_ut/log_sink_etw_2_ut.c
dcristoloveanu: Yes :-).

dcristoloveanu:
Comments got borked. They cannot be resolved anymore.

v2/src/log_sink_etw.c
dcristoloveanu: Good point. Initially it was dynamically allocated, but that would suck. Since we hardcode it, yeah, it can be in the struct with its size.

v2/tests/log_sink_etw_2_ut/log_sink_etw_2_ut.c
```diff
@@ -0,0 +1,802 @@
...
+static bool actual_and_expected_match;
+
+static void setup_mocks(void)
+{
+    expected_call_count = 0;
+    actual_call_count = 0;
+    actual_and_expected_match = true;
+}
+
+int mock_printf(const char* format, ...)
```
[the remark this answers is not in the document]
dcristoloveanu: Actually can't be static as the symbol is referenced in the other translation unit (the one that contains the actual code we want to test).

v2/src/log_sink_etw.c
dcristoloveanu: Added

v2/src/log_sink_etw.c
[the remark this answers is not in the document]
dcristoloveanu: Not at all. I'll optimize it further, separate PR will be incoming if you do not mind.

ID Work Item Type Title State Assigned To Effort Remaining Work Original Estimate Value Area Iteration Path Tags
22185802 Task Optimize log_sink_etw to use TLS (thread local storage) To Do           One

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+
+        // alias to the event bytes
+        SELF_DESCRIBED_EVENT* self_described_event = (SELF_DESCRIBED_EVENT*)_tlgEvent;
+
+        uint8_t* pos = &self_described_event->metadata[0];
+
+        // at this point we filled all the event information, now we need to provide the metadata bytes
+        // first one is the event name, followed by metadata for all the fields
+        // copy event name 
+        (void)memcpy(pos, event_name, event_name_length); pos += event_name_length;
```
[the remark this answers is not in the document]
dcristoloveanu: Should be done with using TLS (thread local storage). That will be so juicy!

ID Work Item Type Title State Assigned To Effort Remaining Work Original Estimate Value Area Iteration Path Tags
22185802 Task Optimize log_sink_etw to use TLS (thread local storage) To Do           One

v2/src/log_sink_etw.c
dcristoloveanu: compiler does it.
Raymond has a nice block where he calls us RISC haters. Which we are. Well, Tracelogging is too I guess.

[https://devblogs.microsoft.com/oldnewthing/20200103-00/?p=103290](https://devblogs.microsoft.com/oldnewthing/20200103-00/?p=103290)

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+
+        // copy the field metadata (name and in type)
+
+        /* Codes_SRS_LOG_SINK_ETW_01_034: [ log_sink_etw.log_sink_log shall fill the event metadata: ]*/
+
+        /* Codes_SRS_LOG_SINK_ETW_01_035: [ The string content (as field name, excluding zero terminator), followed by one byte with the value TlgInANSISTRING. ]*/
+        /* Codes_SRS_LOG_SINK_ETW_01_044: [ Length of the content field name (determined at compile time, excluding zero terminator) + 1. ]*/
+        (void)memcpy(pos, "content", sizeof("content")); pos += sizeof("content");
```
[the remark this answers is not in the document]
dcristoloveanu: I think I want to do the pregenerating of TLS saved structures, but I will do that as separate PR, it deserves its own glory.

v2/src/log_sink_etw.c
dcristoloveanu: Added

v2/src/log_sink_etw.c
dcristoloveanu: Changed

v2/src/log_sink_etw.c
dcristoloveanu: Added an extra check

v2/src/log_sink_etw.c
dcristoloveanu: Haha, added some comments about it

v2/src/log_sink_etw.c
dcristoloveanu: Right, if formatted_message_length >= available_bytes then we'll try again using all the space we have. If that also goes over then that's what we can do. I did add a line making sure we terminate it for the 2nd vsnprintf we do.

v2/src/log_sink_etw.c
dcristoloveanu: We'll reset to not using any properties if we hit this error

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+                }
+                else
+                {
+                    /* Codes_SRS_LOG_SINK_ETW_01_054: [ A string with the property name (including zero terminator) ]*/
+                    (void)memcpy(pos, context_property_value_pairs[i].name, name_length + 1);
+                    pos += name_length + 1;
+                }
+
+                /* Codes_SRS_LOG_SINK_ETW_01_055: [ A byte with the type of property, as follows: ]*/
+                switch (context_property_value_pairs[i].type->get_type())
```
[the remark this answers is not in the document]
dcristoloveanu: For now I will leave it here, otherwise I'd have to push the tlg enums into the context. Maaaaaybe that will come at some point, but likely this won't be our top issue now.

v2/src/log_sink_etw.c
dcristoloveanu: I think I want to address this with having prepopulated structures

v2/src/log_sink_etw.c
dcristoloveanu: Yeah, was thinking of this too while writing the code. As in - maybe the context can be friendlier to tracelogging so that we can memcpy (or not event that).
I'll add a task for this.

v2/src/log_sink_etw.c
```diff
@@ -0,0 +1,497 @@
...
+
+        uint16_t metadata_size_without_properties = (uint16_t)(pos - &self_described_event->metadata[0]);
+
+        if (property_value_count <= LOG_MAX_ETW_PROPERTY_VALUE_PAIR_COUNT)
+        {
+            /* Codes_SRS_LOG_SINK_ETW_01_053: [ For each property in log_context the following shall be added to the event metadata: ]*/
+            for (i = 0; i < property_value_count; i++)
+            {
+                /* Codes_SRS_LOG_SINK_ETW_01_051: [ For each property in log_context, the length of the property name + 1 and one extra byte for the type of the field. ]*/
+                size_t name_length = strlen(context_property_value_pairs[i].name);
```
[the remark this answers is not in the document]
dcristoloveanu: ID Work Item Type Title State Assigned To Effort Remaining Work Original Estimate Value Area Iteration Path Tags
22271413 Task Optimize the context layout to be friendlier to tracelogging To Do           One

v2/src/log_sink_etw.c
dcristoloveanu: ID Work Item Type Title State Assigned To Effort Remaining Work Original Estimate Value Area Iteration Path Tags
22271413 Task Optimize the context layout to be friendlier to tracelogging To Do           One

dcristoloveanu:
> `log_sink_etw` implements a log sink interface that logs the events as ETW events.

I copy/pasted the example and description from TraceLoggingProvider.h. I don't think I want to invent a better one.

---
In reply to: [1544163617](https://github.com/Azure/c-logging/pull/123#issuecomment-1544163617) [](http://example.com/codeflow?ancestors=1544163617,1544163617)

---
Refers to: v2/devdoc/log_sink_etw_requirements.md:3 in 985d73d. [](commit_id = 985d73dffd4f4ebed1afc4f196f1366c7f3c3530, deletion_comment = False)

v2/devdoc/log_sink_etw_requirements.md
dcristoloveanu: I added the description of how an event is assembled in the code.

v2/src/log_sink_etw.c
dcristoloveanu: Made an internal function so that we do not run the risk. We would not really run the risk as we never unregister right now, but better not run the risk at all.

anporumb (approved):
:shipit:

PR #124, opened 2023-05-17
Log sink etw int

dcristoloveanu:
Let's have some log sink etw int tests


v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+static GUID provider_guid = { 0xDAD29F36, 0x0A48, 0x4DEF, { 0x9D, 0x50, 0x8E, 0xF9, 0x03, 0x6B, 0x92, 0xB4 } };
+
+MU_DEFINE_ENUM_STRINGS(LOG_CONTEXT_PROPERTY_TYPE, LOG_CONTEXT_PROPERTY_TYPE_VALUES)
+
+#define POOR_MANS_ASSERT(cond) \
+    if (!(cond)) \
+    { \
+        (void)printf("%s:%d test failed\r\n", __FUNCTION__, __LINE__); \
+        asserts_failed++; \
+        abort(); \
```
anporumb, line 36: >abort [](http://example.com/codeflow?start=8&length=5)

So if this abort() ever executes, do we need asserts_failed too? Seeing as unless in a heavy threaded environment, at most assert_failed will get to is "1".

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+        (void)memset(&parsed_events[current_parsed_event_count], 0, sizeof(PARSED_EVENT));
+
+        EVENT_HEADER_EXTENDED_DATA_ITEM* extended_data_items = pEventRecord->ExtendedData;
+
+        for (unsigned short i = 0; i < pEventRecord->ExtendedDataCount; i++)
+        {
+            if (extended_data_items[i].ExtType == EVENT_HEADER_EXT_TYPE_EVENT_SCHEMA_TL)
+            {
+                // This one has the event name
+                // Not really documented, but the Windows source code seems to indicate I can just skip over a few bytes that hold a length and version
```
anporumb, line 119: >hold a length [](http://example.com/codeflow?start=126&length=13)

Would that be the length of the string? Asking because strncpy below always does more than needed or doesn't copy the whole string...

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+                const char* event_name = (const char*)(uintptr_t)(extended_data_items[i].DataPtr + 4);
+
+                if (event_name != NULL)
+                {
+                    (void)strncpy(parsed_events[current_parsed_event_count].event_name, event_name, sizeof(parsed_events[current_parsed_event_count].event_name));
+                    parsed_events[current_parsed_event_count].event_name[sizeof(parsed_events[current_parsed_event_count].event_name) - 1] = '\0';
+                    metadata = (const uint8_t*)(event_name + strlen(event_name) + 1);
```
anporumb, line 126: >strlen [](http://example.com/codeflow?start=61&length=6)

if we've going to anyway evaluate `strlen` for event_name once, then we might as well use memcpy to copy the string into its "parsed_events" destination.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+                }
+
+                break;
+            }
+        }
+       
+        POOR_MANS_ASSERT(pEventRecord->UserData != NULL);
+
+        const char* current_user_data_field = pEventRecord->UserData;
+        if (current_user_data_field != NULL)
```
anporumb, line 136: >current_user_data_field [](http://example.com/codeflow?start=12&length=23)

why do we need to evaluate this condition when POOR_MANS_ASSERT is just in line 133 above?

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+        {
+            (void)strncpy(parsed_events[current_parsed_event_count].content, current_user_data_field, sizeof(parsed_events[current_parsed_event_count].content));
+            parsed_events[current_parsed_event_count].content[sizeof(parsed_events[current_parsed_event_count].content) - 1] = '\0';
+            current_user_data_field += strlen(current_user_data_field) + 1;
```
anporumb, line 140: >strlen [](http://example.com/codeflow?start=39&length=6)

I see this as a pattern where strncpy is followed by strlen anyway.

Let's grab strlen first, then see if it fits, then memcpy all the way :)

./here and everywhere.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+    (void)log_file;
+
+    return TRUE;
+}
+
+static void WINAPI event_trace_record_callback(EVENT_RECORD* pEventRecord)
+{
+    if (IsEqualGUID(&pEventRecord->EventHeader.ProviderId, &provider_guid))
+    {
+        const uint8_t* metadata = NULL;
```
anporumb, line 107: >metadata [](http://example.com/codeflow?start=23&length=8)

metadata seems to be uint8_t, but it also wants to be used with strcmp.

What would happen when metadata is const char*?

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+
+        const char* current_user_data_field = pEventRecord->UserData;
+        if (current_user_data_field != NULL)
+        {
+            (void)strncpy(parsed_events[current_parsed_event_count].content, current_user_data_field, sizeof(parsed_events[current_parsed_event_count].content));
+            parsed_events[current_parsed_event_count].content[sizeof(parsed_events[current_parsed_event_count].content) - 1] = '\0';
+            current_user_data_field += strlen(current_user_data_field) + 1;
+        }
+        POOR_MANS_ASSERT(strcmp((const char*)metadata, "content") == 0);
+        metadata += sizeof("content") + 1;
```
anporumb, line 143: >+ 1 [](http://example.com/codeflow?start=38&length=3)

rarely do I see sizeof("some string") + 1. It is usually " -1 " to get the length in characters. What's this +1 byte that needs skipping? A comment would be nice for the reviewer :)

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+
+        if (current_user_data_field != NULL)
```
anporumb, line 145: >if (current_user_data_field != NULL) [](http://example.com/codeflow?start=8&length=36)

how does current_user_data_field ever get `NULL` here? overflow on 64bit memory boundary? Maybe there's a different condition (or none at all) that needs to be written here?

anporumb: Pattern detected so "here and everywhere"

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+
+        uint32_t property_index = 0;
+
+        // get all other properties
+        while (current_user_data_field - (const char*)pEventRecord->UserData < pEventRecord->UserDataLength)
+        {
+            // get name and type from metadata
+            (void)strncpy(parsed_events[current_parsed_event_count].properties[property_index].property_name, (const char*)metadata, sizeof(parsed_events[current_parsed_event_count].properties[property_index].property_name));
+            parsed_events[current_parsed_event_count].properties[property_index].property_name[sizeof(parsed_events[current_parsed_event_count].properties[property_index].property_name) - 1] = '\0';
+            metadata += strlen(parsed_events[current_parsed_event_count].properties[property_index].property_name) + 1;
```
anporumb, line 177: >strlen [](http://example.com/codeflow?start=24&length=6)

strlen (here) + strncpy (line 175) != love.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+            parsed_events[current_parsed_event_count].properties[property_index].property_type = *metadata;
+            metadata++;
+
+            switch (parsed_events[current_parsed_event_count].properties[property_index].property_type)
+            {
+            default:
+                POOR_MANS_ASSERT(0);
+                break;
+            case _TlgInSTRUCT | _TlgInChain:
+                parsed_events[current_parsed_event_count].properties[property_index].struct_field_count = *(uint8_t*)metadata;
```
anporumb, line 187: >uint8_t [](http://example.com/codeflow?start=108&length=7)

this cast doesn't need to lose the const qualifier.

See it as a pattern, so "here and everywhere".

Also, why is it necessary at all when metadata is "const uint8_t*". I know this is contradicting another comment of mine (phrased as a question) from before.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+            case TlgInUINT8:
+                parsed_events[current_parsed_event_count].properties[property_index].uint8_t_value = *(uint8_t*)current_user_data_field;
+                current_user_data_field += sizeof(uint8_t);
+                break;
+            case TlgInINT8:
+                parsed_events[current_parsed_event_count].properties[property_index].int8_t_value = *(int8_t*)current_user_data_field;
+                current_user_data_field += sizeof(int8_t);
+                break;
+            case TlgInUINT16:
+                parsed_events[current_parsed_event_count].properties[property_index].uint16_t_value = *(uint16_t*)current_user_data_field;
```
anporumb, line 204: >uint16_t [](http://example.com/codeflow?start=104&length=8)

I don't think this is right.

`current_user_data_field` is not guaranteed to be aligned (at all). Casting and reading unaligned works wonder on Intel platforms, on others... less so.

Let's use some primitives that read byte by byte and assemble the bytes in "values" after the bytes are read.

Also, we can only hope that the endianess of the bytes matches what we expect as a platform (different concern).

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+        }
+
+        parsed_events[current_parsed_event_count].property_count = property_index;
+
+        (void)InterlockedIncrement(&parsed_event_count);
+        WakeByAddressAll((void*)&parsed_event_count);
+    }
+}
+
+static char test_path[MAX_PATH];
```
anporumb, line 239: >test_path [](http://example.com/codeflow?start=12&length=9)

Is thsi really used? Seems to be filled with some temp path later and then nothing happens to it.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+
+static void generate_trace_session(void)
+{
+    UUID uuid;
+    POOR_MANS_ASSERT(UuidCreate(&uuid) == RPC_S_OK);
+    int snprintf_result = snprintf(trace_session_name, sizeof(trace_session_name), "%s-%08X-%04X-%04X-%02X%02X-%02X%02X%02X%02X%02X%02X",
+        __FUNCTION__, uuid.Data1, uuid.Data2, uuid.Data3, uuid.Data4[0], uuid.Data4[1], uuid.Data4[2], uuid.Data4[3], uuid.Data4[4], uuid.Data4[5], uuid.Data4[6], uuid.Data4[7]);
```
anporumb, line 246: >__FUNCTION__ [](http://example.com/codeflow?start=8&length=12)

so here `__FUNCTION__` is always expanded to `generate_trace_session`, and I have to wonder if that's intended or has the code been moved with copy&paste from some other place where `__FUNCTION__` was different every time :)

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+    POOR_MANS_ASSERT((snprintf_result > 0) && (snprintf_result < sizeof(trace_session_name)));
+}
+
+static TRACEHANDLE start_trace(uint8_t trace_level)
+{
+    // setup the etw consumer
+    TRACEHANDLE trace_session_handle;
+
+    (void)printf("Starting trace session %s\r\n", trace_session_name);
```
anporumb, line 255: >trace_session_name [](http://example.com/codeflow?start=50&length=18)

I did not follow the complete code when I am making this comment, but who guarantees that at this moment "trace_session_name" is even `'\0'` terminated? I don't see any ` = { 0 }` where `trace_session_name` is defined.

anporumb: Sort of "consider initializing all static globals" here and everywhere.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+    start_event_trace_property_data.props.MinimumBuffers = 4;
+    start_event_trace_property_data.props.MaximumBuffers = 4;
+    start_event_trace_property_data.props.BufferSize = 64 * 1024;
+    start_event_trace_property_data.props.FlushTimer = 1;
+    start_event_trace_property_data.props.LogFileNameOffset = 0;
+    start_event_trace_property_data.props.LoggerNameOffset = offsetof(EVENT_TRACE_PROPERTY_DATA, logger_name);
+
+    POOR_MANS_ASSERT(StartTraceA(&trace_session_handle, trace_session_name, &start_event_trace_property_data.props) == ERROR_SUCCESS);
+
+    (void)strcpy(logger_name, start_event_trace_property_data.logger_name);
```
anporumb, line 271: >strcpy [](http://example.com/codeflow?start=10&length=6)

Should check that `start_event_trace_property_data.logger_name` fits in 128 bytes declared above.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+static TRACEHANDLE start_trace(uint8_t trace_level)
+{
+    // setup the etw consumer
+    TRACEHANDLE trace_session_handle;
+
+    (void)printf("Starting trace session %s\r\n", trace_session_name);
+
+    EVENT_TRACE_PROPERTY_DATA start_event_trace_property_data = { 0 };
+
+    start_event_trace_property_data.props.Wnode.BufferSize = sizeof(start_event_trace_property_data);
```
anporumb, line 259: >Wnode [](http://example.com/codeflow?start=42&length=5)

This WNODE structure has a lot of other information in it ([https://learn.microsoft.com/en-us/windows/win32/etw/wnode-header](https://learn.microsoft.com/en-us/windows/win32/etw/wnode-header)) but I will assume these are the only ones needed for the purpose here.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+} PARSED_EVENT;
+
+#define MAX_EVENTS 128
+
+static volatile LONG parsed_event_count;
+static PARSED_EVENT parsed_events[MAX_EVENTS];
+static HANDLE process_thread_handle;
+static TRACEHANDLE trace;
+static char logger_name[128];
+static char trace_session_name[128];
```
anporumb, line 92: >trace_session_name [](http://example.com/codeflow?start=12&length=18)

Any way we can avoid this (and the others) being global?

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+    stop_event_trace_property_data.props.Wnode.BufferSize = sizeof(stop_event_trace_property_data);
+    stop_event_trace_property_data.props.Wnode.Flags = WNODE_FLAG_TRACED_GUID;
+    stop_event_trace_property_data.props.LoggerNameOffset = offsetof(EVENT_TRACE_PROPERTY_DATA, logger_name);
+    stop_event_trace_property_data.props.LogFileNameOffset = 0;
+
+    (void)printf("Stopping trace session %s\r\n", trace_session_name);
+    POOR_MANS_ASSERT(StopTraceA(trace_session_handle, NULL, &stop_event_trace_property_data.props) == ERROR_SUCCESS);
+}
+
+DWORD WINAPI process_trace_thread_func(void* arg)
```
anporumb, line 295: >process_trace_thread_func [](http://example.com/codeflow?start=13&length=25)

static?

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+    parsed_event_count = 0;
+
+    log_file.ProcessTraceMode = PROCESS_TRACE_MODE_REAL_TIME | PROCESS_TRACE_MODE_EVENT_RECORD;
+    log_file.EventRecordCallback = event_trace_record_callback;
+    log_file.BufferCallback = event_trace_buffer_callback;
+    log_file.LoggerName = logger_name;
+    log_file.Context = NULL;
+
+    trace = OpenTraceA(&log_file);
+    POOR_MANS_ASSERT(trace != INVALID_PROCESSTRACE_HANDLE);
```
anporumb, line 322: >trace [](http://example.com/codeflow?start=21&length=5)

looks like "trace" should be a parameter passed to process_trace_thread_func. Just saying that there's a bit too many globals here.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+        if (current_parsed_event_count == expected_event_count)
+        {
+            break;
+        }
+
+        (void)WaitOnAddress(&parsed_event_count, &current_parsed_event_count, sizeof(int32_t), INFINITE);
+    } while (1);
+}
+
+static void log_sink_etw_log_with_LOG_LEVEL_ERROR_succeeds(void)
```
anporumb, line 354: >log_sink_etw_log_with_LOG_LEVEL_ERROR_succeeds [](http://example.com/codeflow?start=12&length=46)

Looks like this is "special" because it captures also the initial setup "self test" of the logging provider.

(as opposed to log_sink_etw_log_all_levels_when_all_levels_enabled_succeeds).

So that would imply that changing the order in which `log_sink_etw_log_with_LOG_LEVEL_ERROR_succeeds` and `log_sink_etw_log_all_levels_when_all_levels_enabled_succeeds` are written would lead to tests failing.

So I would propose to have a "eat_first_event" with some description about why it is discarded. Had to uncompile lines 376-381

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+            // act
+            log_sink_etw.log_sink_log(event_log_level, NULL, __FILE__, __FUNCTION__, captured_line, "test_event");
+
+            // assert
+            wait_for_event_count(expected_event_count);
+
+            // if we are not expecting an event, wait for some time just to make sure no event is seen
+            if (expected_event_count == 0)
+            {
+                (void)printf("Waiting for %.02f to make sure no events are published ...", (double)TEST_WAITTIME_NO_EVENT / 1000);
```
anporumb, line 478: >Waiting [](http://example.com/codeflow?start=30&length=7)

nice

anporumb, line 478: >%.02f [](http://example.com/codeflow?start=42&length=5)

this should indicate what the wait is measured in (in this case, seems to be seconds).

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+    LOG_CONTEXT_CREATE(log_context, NULL,
+        LOG_CONTEXT_STRING_PROPERTY(gigi, "duru"),
+        LOG_CONTEXT_PROPERTY(int8_t, prop1, 42),
+        LOG_CONTEXT_PROPERTY(uint8_t, prop2, 43),
+        LOG_CONTEXT_PROPERTY(int16_t, prop3, 44),
+        LOG_CONTEXT_PROPERTY(uint16_t, prop4, 45),
+        LOG_CONTEXT_PROPERTY(int32_t, prop5, 46),
+        LOG_CONTEXT_PROPERTY(uint32_t, prop6, 47),
+        LOG_CONTEXT_PROPERTY(int64_t, prop7, 48),
+        LOG_CONTEXT_PROPERTY(uint64_t, prop8, 49)
```
anporumb, line 542: >uint64_t [](http://example.com/codeflow?start=29&length=8)

due to how the values are picked from memory, may I suggest values that fill all the bytes of the type, with different values?

In this case, a 0x12345678ABCDEF01 value would fill all the 8 bytes with different values.

anporumb: For all types, here and everywhere.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+    // arrange
+    generate_trace_session();
+
+    TRACEHANDLE trace_session_handle = start_trace(TRACE_LEVEL_VERBOSE);
+
+    int captured_line = __LINE__;
+
+    LOG_CONTEXT_HANDLE log_context;
+    LOG_CONTEXT_CREATE(log_context, NULL,
+        LOG_CONTEXT_STRING_PROPERTY(gigi, "duru"),
```
anporumb, line 534: >gigi, "duru" [](http://example.com/codeflow?start=36&length=12)

A wild mythic creature suddenly appears!

anporumb (approved):
:shipit:

v2/tests/log_sink_etw_int/log_sink_etw_int.c
jebrando, line 239: nitpick: This variable just seems randomly placed in the file, I was looking for it when I was going through main and couldn't find it in the usual places.  Could you please move it up near the top with the other statics, or down to main where it's used.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+        log_sink_etw_log_each_individual_level();
+        log_sink_etw_log_with_context_with_properties();
+        log_sink_etw_log_with_context_with_nested_structs();
+
+        result = (int)asserts_failed;
+    }
+
+    return result;
+}
+
```
jebrando, line 683: nitpick: remove unnecessary line

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+    process_thread_handle = CreateThread(NULL, 0, process_trace_thread_func, NULL, 0, &thread_id);
+    POOR_MANS_ASSERT(process_thread_handle != NULL);
+
+    // Give the thread some time to start
+    Sleep(TEST_PARSE_START_DELAY);
+}
+
+static void stop_parse_events(void)
+{
+    WaitForSingleObject(process_thread_handle, INFINITE);
```
jebrando, line 334: >WaitForSingleObject [](http://example.com/codeflow?start=4&length=19)

(void)

jebrando (approved):
:shipit:

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: We don't need the count of asserts any longer, I will remove it from this test suite and have a task to clean the other test suites in this repo.

dcristoloveanu: Actually changed it in all the test suites in this repo.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+    return TRUE;
+}
+
+static void WINAPI event_trace_record_callback(EVENT_RECORD* pEventRecord)
+{
+    if (IsEqualGUID(&pEventRecord->EventHeader.ProviderId, &provider_guid))
+    {
+        const uint8_t* metadata = NULL;
+
+        LONG current_parsed_event_count = InterlockedAdd(&parsed_event_count, 0);
```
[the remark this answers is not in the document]
dcristoloveanu: ewww, yes, changed to having the parsed_event_count be initialized to -1. And then using InterlockedIncrement here to grab the slot in the array. ewwww, again, as I just realized that the parsed_event_count was actually initialized without Interlocked. bleh.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Changed to having an extracted function that copies the strings and uses just memcpy. Also less code, so t is nicer now.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: This comment is from the original authors (Matt). I believe it refers to the internal structures of TraceLogging. I shall change the comment accordingly.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Not needed, removed.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Done in all places by extracting a function.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: We'll have to cast one way or the other. So I'll leave it as is.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Added comment.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Actually made metadata be const char*. Seems to make for less casts.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Fixed.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Added some local read functions for this

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Leftover from the times of logging in a file.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Was moved when I extracted the setup and helpers. Corrected now to take a function_name argument.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: We have a POOR_MANS_ASSERT with abort in it and we assert on the return of snprintf.
That should be enough per C11

Thus, the null-terminated output has been
completely written if and only if the returned value is nonnegative and less than n.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Added ASSERT

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: I assume the same.  started from a sample that only had this.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: static-ized.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
```diff
@@ -0,0 +1,683 @@
...
+
+    trace = OpenTraceA(&log_file);
+    POOR_MANS_ASSERT(trace != INVALID_PROCESSTRACE_HANDLE);
+
+    DWORD thread_id;
+    process_thread_handle = CreateThread(NULL, 0, process_trace_thread_func, NULL, 0, &thread_id);
+    POOR_MANS_ASSERT(process_thread_handle != NULL);
+
+    // Give the thread some time to start
+    Sleep(TEST_PARSE_START_DELAY);
```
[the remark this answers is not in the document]
dcristoloveanu: We miss events and there is some level of flakiness without a Sleep. I also considered starting the thread and making it hang in a tight interlocked loop. But for now this seems to be good enough.

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: I added an extra comment. I wanted to also check this self test event rather than just discard it.,

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Added

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: :-)

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Removed

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Added POOR_MANS_ASSERT actually

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Removed

v2/tests/log_sink_etw_int/log_sink_etw_int.c
dcristoloveanu: Done.

PR #129, opened 2023-05-31
link with logging_stacktrace.c to enable log stack in bs_watchdog

M-iceberg:


CMakeLists.txt
```diff
@@ -84,16 +84,17 @@ if(${use_etw} STREQUAL "TRACELOGGING")
     add_definitions(-DUSE_TRACELOGGING)
 endif() #(${use_etw} STREQUAL "TRACELOGGING")
```
mattdurak, line 86: Can you indent everything inside the `if(WIN32)` and else blocks? Was a bit confused at first because I missed the nested if's

mattdurak (approved):
:shipit:

PR #140, opened 2023-06-16
Add a warning suppress

dcristoloveanu:
Add a warning suppress

### DIFF
--- a/v2/inc/c_logging/log_context.h
+++ b/v2/inc/c_logging/log_context.h
@@ -47,6 +47,7 @@ const LOG_CONTEXT_PROPERTY_VALUE_PAIR* log_context_get_property_value_pairs(LOG_
 int internal_log_context_init_from_parent(LOG_CONTEXT_HANDLE dest_log_context, LOG_CONTEXT_HANDLE parent_log_context);
 uint32_t internal_log_context_get_values_data_length_or_zero(LOG_CONTEXT_HANDLE log_context);
 
+MU_SUPPRESS_WARNING(4505) // warning C4505: 'internal_log_context_get_property_value_pair_count_or_zero': unreferenced function with internal linkage has been removed
 static uint32_t internal_log_context_get_property_value_pair_count_or_zero(LOG_CONTEXT_HANDLE log_context)
 {
     uint32_t result = log_context_get_property_value_pair_count(log_context);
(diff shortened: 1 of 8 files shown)


jasmineymlo (approved):

v2/inc/c_logging/log_context.h
```diff
@@ -47,6 +47,7 @@ const LOG_CONTEXT_PROPERTY_VALUE_PAIR* log_context_get_property_value_pairs(LOG_
 int internal_log_context_init_from_parent(LOG_CONTEXT_HANDLE dest_log_context, LOG_CONTEXT_HANDLE parent_log_context);
 uint32_t internal_log_context_get_values_data_length_or_zero(LOG_CONTEXT_HANDLE log_context);
 
+MU_SUPPRESS_WARNING(4505) // warning C4505: 'internal_log_context_get_property_value_pair_count_or_zero': unreferenced function with internal linkage has been removed
```
anporumb, line 50: >4505 [](http://example.com/codeflow?start=20&length=4)

So what is Linux throwing here? Same error number ./doubt. This doesn't seem to be necessarily ETW related so...

v2/inc/c_logging/log_context.h
```diff
@@ -47,6 +47,7 @@ const LOG_CONTEXT_PROPERTY_VALUE_PAIR* log_context_get_property_value_pairs(LOG_
...
 static uint32_t internal_log_context_get_property_value_pair_count_or_zero(LOG_CONTEXT_HANDLE log_context)
```
anporumb, line 51: >static [](http://example.com/codeflow?start=0&length=6)

why does this need to be static [and have code].

Can it be moved to its own object file (separate .c/.h) and then only included on a "needed" basis?

anporumb, line 51: >internal_log_context_get_property_value_pair_count_or_zero [](http://example.com/codeflow?start=16&length=58)

this whole function seems to exist because it wants to switch the result of `log_context_get_property_value_pair_count` (note: public API) from UINT32_MAX to 0. 

What would happen : 
if "1" :`log_context_get_property_value_pair_count` would return "0" when it returns UINT32_MAX?

if "2": whoever is calling now `internal_log_context_get_property_value_pair_count_or_zero` sould actually call `log_context_get_property_value_pair_count` and examine the return value itself?

anporumb (approved):
:shipit:

v2/inc/c_logging/log_context.h
dcristoloveanu: Apparently Linux did not throw anything ... yet, likely because I hit this when I pulled it in the ebs repo. Maybe another one will be coming for Linux when I hit it :-).

v2/inc/c_logging/log_context.h
dcristoloveanu: A very good point, I shall try that

dcristoloveanu: Well, I think I will remove this and make log_context_get_property_value_pair_count just return 0. Makes it easier.

dcristoloveanu: So, did this one. Makes everything simpler.

PR #146, opened 2023-07-05
add wchar string type

M-iceberg:


v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
```diff
@@ -0,0 +1,95 @@
+# `log_context_property_type_wchar_t_ptr` requirements
+
+`log_context_property_type_wchar_t_ptr` implements the code needed for handling an wchar_t string.
```
anporumb, line 3: >an [](http://example.com/codeflow?start=80&length=2)

extraneous "n"

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
```diff
@@ -0,0 +1,95 @@
...
+
+`LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string` produces the string representation of a `wchar_t_ptr`.
+
+**SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_001: [** If `property_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string` shall fail and return a negative value.  **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_002: [** If `buffer` is `NULL` and `buffer_length` is greater than 0, `LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string` shall fail and return a negative value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_003: [** If `buffer` is `NULL` and `buffer_length` is 0, `LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string` shall return the length of the `wchar_t` string pointed to by `property_value`. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_004: [** Otherwise, `LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string` shall copy the wchar_t string pointed to by `property_value` to `buffer` by using `snprintf` with `buffer`, `buffer_length` and format string `%ls` and pass in the values list the `const wchar_t*` value pointed to be `property_value`. **]**
```
anporumb, line 29: >be [](http://example.com/codeflow?start=340&length=2)

by

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
```diff
@@ -0,0 +1,95 @@
...
+
+**SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_005: [** `LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string` shall succeed and return the result of `snprintf`. **]**
```
anporumb, line 31: >return the result of `snprintf` [](http://example.com/codeflow?start=133&length=31)

This is interesting because snprintf can (and will) return values greater than `buffer_length` in truncation cases.

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
```diff
@@ -0,0 +1,95 @@
...
+typedef int (*LOG_CONTEXT_PROPERTY_TYPE_COPY)(void* dst_value, const void* src_value);
+```
+
+`LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).copy` copies the `wchar_t_ptr` value from the address pointed by `src_value` to `dst_value`.
+
+**SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_007: [** If `src_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_008: [** If `dst_value` is `NULL`, `LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).copy` shall fail and return a non-zero value. **]**
+
+**SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_009: [** `LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).copy` shall copy the entire string (including the null terminator) from `src_value` to `dst_value`. **]**
```
anporumb, line 47: >copy [](http://example.com/codeflow?start=116&length=4)

Just wondering at this moment what are the guarantees that `dst_value` has enough memory to accomodate `src_value`.

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
```diff
@@ -0,0 +1,95 @@
...
+
+## LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr)
+
+```c
+int LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr)(const char* format, ...);
+```
+
+`LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr)` returns the amount of memory needed in number of wide-characters to store the `printf` style formatted string given by `format` and the arguments in `...`.
+
+**SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_018: [** If `_vsnwprintf` fails, `LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr)` shall return a negative value. **]**
```
anporumb, line 93: >_ [](http://example.com/codeflow?start=60&length=1)

is this `_` here intended?

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
```diff
@@ -0,0 +1,95 @@
...
+
+**SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_019: [** Otherwise, on success, `LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr)` shall return the amount of memory needed in number of wide-chracters to store the `wprintf` style formatted `wchar_t` string given by `format` and the arguments in `...`. **]**
```
anporumb, line 95: >wprintf [](http://example.com/codeflow?start=222&length=7)

`wprintf` needs `wchar_t` argument. 

Line 88 lists `const char* format`. Which one is correct?

v2/inc/c_logging/log_context.h
```diff
@@ -28,6 +28,7 @@ extern "C" {
 
 #define LOG_MAX_STACK_DATA_SIZE                 4096
 #define LOG_MAX_STACK_PROPERTY_VALUE_PAIR_COUNT 64
+#define LOG_MAX_WCHAR_STRING_LENGTH 512
```
anporumb, line 32: >512 [](http://example.com/codeflow?start=36&length=3)

Just saying that in the world of our logging this size might be not enough.

I've encountered strings > 4K while doing certificates, simply due to the amount of information needed there.

What are the implications of making this 4k, let's say?

v2/inc/c_logging/log_context.h
```diff
@@ -110,6 +116,15 @@ uint32_t internal_log_context_get_values_data_length_or_zero(LOG_CONTEXT_HANDLE
     data_pos += sprintf(property_value_pair->value, __VA_ARGS__) + 1; \
     property_value_pair++; \
 
+#define EXPAND_SETUP_PROPERTY_PAIR_LOG_CONTEXT_WSTRING_PROPERTY(property_name, ...) \
+    /* Codes_SRS_LOG_CONTEXT_07_001: [ LOG_CONTEXT_WSTRING_PROPERTY shall expand to code allocating a property/value pair of type wchar_t_ptr and the name property_name. ]*/ \
+    property_value_pair->value = data_pos; \
+    property_value_pair->name = MU_TOSTRING(property_name); \
+    property_value_pair->type = &wchar_t_ptr##_log_context_property_type; \
```
anporumb, line 124: >wchar_t_ptr##_log_context_property_type; [](http://example.com/codeflow?start=33&length=40)

Mu_C2?

v2/inc/c_logging/log_context_property_type_wchar_t_ptr.h
```diff
@@ -0,0 +1,22 @@
...
+#include "c_logging/log_context_property_type_if.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    int LOG_CONTEXT_PROPERTY_TYPE_INIT(wchar_t_ptr)(void* dst_value, size_t count, const wchar_t* format, ...);
+    int LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr)(const wchar_t* format, ...);
+
+    extern const LOG_CONTEXT_PROPERTY_TYPE_IF LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr);
```
anporumb, line 16: >extern [](http://example.com/codeflow?start=4&length=6)

why is "extern" needed here? in C all declarations are `extern` default.

v2/inc/c_logging/log_context_property_type_wchar_t_ptr.h
```diff
@@ -0,0 +1,22 @@
...
+#define LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_H
+
+#include "c_logging/log_context_property_type_if.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    int LOG_CONTEXT_PROPERTY_TYPE_INIT(wchar_t_ptr)(void* dst_value, size_t count, const wchar_t* format, ...);
+    int LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr)(const wchar_t* format, ...);
```
anporumb, line 14: >wchar_t [](http://example.com/codeflow?start=72&length=7)

`wchar_t` in C comes from "some header" - which is missing as #include here.

v2/src/log_context_property_type_wchar_t_ptr.c
```diff
@@ -0,0 +1,143 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdarg.h>
+#include <stddef.h>
+#include <stdio.h>
+#include <wchar.h>
+#include <corecrt_wstdio.h>
```
anporumb: did this even compile in Linux? Looks like Windows specific file.

v2/src/log_context_property_type_wchar_t_ptr.c
```diff
@@ -0,0 +1,143 @@
...
+        result = -1;
+    }
+    else
+    {
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_003: [ If buffer is NULL and buffer_length is 0, LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string shall return the length of the wchar_t string pointed to by property_value. ]*/
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_004: [ Otherwise, LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string shall copy the wchar_t string pointed to by property_value to buffer by using snprintf with buffer, buffer_length and format string %ls and pass in the values list the const wchar_t* value pointed to be property_value. ]*/
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_005: [ LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string shall succeed and return the result of snprintf. ]*/
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_006: [ If any error is encountered (truncation is not an error), LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string shall fail and return a negative value. ]*/
+        result = snprintf(buffer, buffer_length, "%ls", (const wchar_t*)property_value);
```
anporumb, line 39: >(const wchar_t*) [](http://example.com/codeflow?start=56&length=16)

why is this cast necessary?

anporumb, line 39: >snprintf [](http://example.com/codeflow?start=17&length=8)

what happens if we use `wcstombs` here?

v2/src/log_context_property_type_wchar_t_ptr.c
```diff
@@ -0,0 +1,143 @@
...
+    int result;
+
+    if (
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_007: [ If src_value is NULL, LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).copy shall fail and return a non-zero value. ]*/
+        (dst_value == NULL) ||
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_008: [ If dst_value is NULL, LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).copy shall fail and return a non-zero value. ]*/
+        (src_value == NULL)
+        )
+    {
+        (void)printf("Invalid arguments: void* dst_value=%p, void* src_value=%p\r\n",
```
anporumb, line 55: >void [](http://example.com/codeflow?start=61&length=4)

parameter is `const` in line 42.

v2/src/log_context_property_type_wchar_t_ptr.c
```diff
@@ -0,0 +1,143 @@
...
+    int result;
+
+    if (
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_013: [ If dst_value is NULL, LOG_CONTEXT_PROPERTY_TYPE_INIT(wchar_t_ptr) shall fail and return a non-zero value. ]*/
+        (dst_value == NULL) ||
+        /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_014: [ If format is NULL, LOG_CONTEXT_PROPERTY_TYPE_INIT(wchar_t_ptr) shall fail and return a non-zero value. ]*/
+        (format == NULL)
+        )
+    {
+        (void)printf("Invalid arguments: void* dst_value=%p, const wchar_t* format=%p\r\n",
```
anporumb, line 93: >%p [](http://example.com/codeflow?start=83&length=2)

MU_WP_OR_NULL?

v2/src/log_context_property_type_wchar_t_ptr.c
```diff
@@ -0,0 +1,143 @@
...
+}
+
+int LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr)(const wchar_t* format, ...)
+{
+    va_list args;
+    va_start(args, format);
+
+    /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_018: [ If _vsnwprintf fails, LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr) shall return a negative value. ]*/
+    /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_019: [ Otherwise, on success, LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr) shall return the amount of memory needed in number of wide-chracters to store the wprintf style formatted wchar_t string given by format and the arguments in .... ]*/
+    int result = _vsnwprintf(NULL, 0, format, args) * 2;
```
anporumb: >2 [](http://example.com/codeflow?start=54&length=1)

sizeof(wchar_t)

v2/src/log_context_property_type_wchar_t_ptr.c
```diff
@@ -0,0 +1,143 @@
...
+    if (result >= 0)
+    {
+        result+=2;
```
anporumb, line 133: >2 [](http://example.com/codeflow?start=16&length=1)

sizeof(wchar_t)

v2/src/log_context_property_type_wchar_t_ptr.c
anporumb: >_vsnwprintf [](http://example.com/codeflow?start=17&length=11)

what happens if we use `vswprintf` here?

v2/inc/c_logging/log_context_property_type_wchar_t_ptr.h
anporumb: Ok, now I understand and I like it with `extern`.

I like the explicitness.

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
```diff
@@ -0,0 +1,254 @@
...
+
+static void wchar_t_ptr_to_string_with_NULL_buffer_and_zero_buffer_length_returns_string_length(void)
+{
+    // arrange
+
+    // act
+    int result = LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string(L"cucu", NULL, 0);
+
+    // assert
+    POOR_MANS_ASSERT(result == (int)strlen("cucu"));
```
anporumb, line 54: >strlen [](http://example.com/codeflow?start=36&length=6)

wcslen(L"cucu")

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
```diff
@@ -0,0 +1,254 @@
...
+{
+    // arrange
+    char buffer[4];
+
+    // act
+    int result = LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).to_string(L"bau", buffer, sizeof(buffer));
+
+    // assert
+    POOR_MANS_ASSERT(result == 3);
+    POOR_MANS_ASSERT(strcmp(buffer, "bau") == 0);
```
anporumb, line 67: >strcmp [](http://example.com/codeflow?start=21&length=6)

let's use the wchar_t function for this.

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
```diff
@@ -0,0 +1,254 @@
...
+    POOR_MANS_ASSERT(wcscmp(buffer, L"cucu") == 0);
+}
+
+static void wchar_t_ptr_LOG_CONTEXT_PROPERTY_TYPE_INIT_with_multiple_args_succeeds(void)
+{
+    // arrange
+    wchar_t buffer[43];
+
+    // act
+    int result = LOG_CONTEXT_PROPERTY_TYPE_INIT(wchar_t_ptr)(buffer, 50, L"The answer is %d and let's say hello %ls", 42, L"world");
```
anporumb, line 198: >50 [](http://example.com/codeflow?start=69&length=2)

isn't this 50 wrong here? Seeing as line 194 contains "43".

Consider using sizeof(buffer)/sizeof(buffer[0]) to compute the number of elements in buffer array?

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
```diff
@@ -0,0 +1,254 @@
...
+
+static void wchar_t_ptr_LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE_with_multiple_args_succeeds(void)
+{
+    // arrange
+
+    // act
+    int result = LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr)(L"The answer is %d and let's say hello %ls", 43, L"world");
+
+    // assert
+    POOR_MANS_ASSERT(result == 43);
```
anporumb: >43 [](http://example.com/codeflow?start=31&length=2)

why is this 43 and not 39?

anporumb: I've counted (with chatbot) as 39 the number of characters...

v2/tests/log_context_property_type_wchar_t_ptr_ut/log_context_property_type_wchar_t_ptr_mocked.c
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdio.h>
```
anporumb, line 4: >stdio [](http://example.com/codeflow?start=10&length=5)

likely not needed

anporumb:
> #include "log_sink_etw.c"

... and then I heard that it doesn't build in the gate, so I will be waiting until 
1) comments here are addressed
2) the gate builds

ping me when I need to look at it again.

---
Refers to: v2/tests/log_sink_etw_ut/log_sink_etw_mocked.c:41 in 0ae1072. [](commit_id = 0ae1072057710a67eca9b455b8188bbe1ba23cf0, deletion_comment = False)

anporumb (changes requested):
:clock1:

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
M-iceberg: changed to use wcstombs, in case of length limit forces a stop, number of bytes written is returned. maximum value can be returned is buffer_length

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
M-iceberg: no guarantees, change to use wcscpy_s here and use strcpy_s in ascii_char_ptr_log_context_property_type_copy

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
M-iceberg: removed

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
M-iceberg: done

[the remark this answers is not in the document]
M-iceberg: snprintf doesn't work for linux, modified to use wcstombs here. 

ascii_char_ptr_to_string also use snprintf, the copying of the string can be somewhat optimized by not using `snprintf`, this will be considered in a separate task.

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
M-iceberg: originally use _vsnwprintf, but it's ms specific and can't compile on gate. changed to use vswprintf here

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
M-iceberg: wchar_t is correct, modified line 88

v2/inc/c_logging/log_context.h
M-iceberg: changed to use 4k here.

In log_context_property_type_wchar_t_ptr.c line 127, allocated sizeof LOG_MAX_WCHAR_STRING_LENGTH to get the initial data size but not actually use the buffer and free right after get the data size in line 130. I was wondering if there would be some overhead if allocated more memory then needed here.

v2/inc/c_logging/log_context.h
```diff
@@ -110,6 +116,15 @@ uint32_t internal_log_context_get_values_data_length_or_zero(LOG_CONTEXT_HANDLE
...
+    /* Codes_SRS_LOG_CONTEXT_07_002: [ LOG_CONTEXT_WSTRING_PROPERTY shall expand to code that stores as value a wchar string that is constructed using wprintf-like formatting based on format and all the arguments in .... ]*/ \
+    data_pos += swprintf(property_value_pair->value, LOG_MAX_WCHAR_STRING_LENGTH, __VA_ARGS__)*2 + 2; \
```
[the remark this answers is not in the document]
M-iceberg: done

v2/inc/c_logging/log_context_property_type_wchar_t_ptr.h
```diff
@@ -0,0 +1,22 @@
...
+
+    extern const LOG_CONTEXT_PROPERTY_TYPE_IF LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_H */
```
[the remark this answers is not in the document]
M-iceberg: done

v2/inc/c_logging/log_context_property_type_wchar_t_ptr.h
M-iceberg: added #include <wchar.h>

v2/src/log_context_property_type_wchar_t_ptr.c
M-iceberg: doesn't compile on linux, deleted

v2/src/log_context_property_type_wchar_t_ptr.c
M-iceberg: modified to use wcstombs here

M-iceberg: removed

v2/src/log_context_property_type_wchar_t_ptr.c
M-iceberg: added const

v2/src/log_context_property_type_wchar_t_ptr.c
M-iceberg: modified

v2/src/log_context_property_type_wchar_t_ptr.c
M-iceberg: done

M-iceberg: yes, modified to use vswprintf now

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
M-iceberg: done

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
M-iceberg: buffer has type char* here, so use strcmp, wcscmp can't compile

v2/tests/log_context_property_type_wchar_t_ptr_ut/log_context_property_type_wchar_t_ptr_mocked.c
M-iceberg: removed

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
M-iceberg: size is 43 units of wide char, changed to 43*sizeof(wchar_t)

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
M-iceberg: modified to use sizeof(buffer)/sizeof(buffer[0])

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
```diff
@@ -0,0 +1,254 @@
...
+    POOR_MANS_ASSERT(result != 0);
+}
+
+static void wchar_t_ptr_LOG_CONTEXT_PROPERTY_TYPE_INIT_with_NULL_format_fails(void)
+{
+    // arrange
+    wchar_t buffer[5];
+
+    // act
+    int result = LOG_CONTEXT_PROPERTY_TYPE_INIT(wchar_t_ptr)(buffer, 1, NULL);
```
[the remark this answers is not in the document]
M-iceberg: done

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
```diff
@@ -0,0 +1,254 @@
...
+    int result = LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).copy(buffer, NULL);
+
+    // assert
+    POOR_MANS_ASSERT(result != 0);
+}
+
+static void wchar_t_ptr_copy_with_NULL_dst_value_fails(void)
+{
+    // arrange
+    wchar_t buffer[128];
```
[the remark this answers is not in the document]
M-iceberg: done

v2/src/log_context_property_type_wchar_t_ptr.c
M-iceberg: modified

v2/inc/c_logging/log_context_property_type_wchar_t_ptr.h
M-iceberg: I don't understand why I paste it. Why do we need extern here?

v2/inc/c_logging/log_context.h
M-iceberg: for some reason which I don't understand, MU_C2 doesn't work here

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
M-iceberg: strcpy_s/wcscpy_s not  supported on linux, don't know how to make it safe

v2/inc/c_logging/log_context.h
anporumb: Errr... logging absolutely cannot use any memory allocation.

v2/src/log_context_property_type_wchar_t_ptr.c
```diff
@@ -0,0 +1,151 @@
...
+
+    return result;
+}
+
+int LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr)(const wchar_t* format, ...)
+{
+    va_list args;
+    va_start(args, format);
+
+    wchar_t* buffer = (wchar_t*)malloc(sizeof(wchar_t)* LOG_MAX_WCHAR_STRING_LENGTH);
```
anporumb, line 127: >malloc [](http://example.com/codeflow?start=32&length=6)

Cannot be used, make it a static array here.

v2/inc/c_logging/log_context.h
anporumb: Also: doesn't seem changed (I still see it as 512)

anporumb: Also: seems to be defined in 2 places, which is the source of truth?

LOG_MAX_WCHAR_STRING_LENGTH in log_context.h:33 and in log_context_property_type_wchar_t_ptr.c:17

v2/inc/c_logging/log_context.h
anporumb: Try using MU_C2A... see what happens if it works. Totally wild guess.

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
anporumb: well - if there are "no guarantees" then it might overflow and overwrite some other parts of memory. 

Who calls this function... must have prepared "dst_value" somehow. So if there are any guarantees, they come from the preparer of `dst_value`. 

Can you please trace from where is dst_value coming and see if its memory size matches src_value?

v2/inc/c_logging/log_context_property_type_wchar_t_ptr.h
anporumb: We don't "need"; it would without "extern" equally nice.

But I like it "with" extern - as it makes the intention very clear. We usually don't have extern variables (note:  LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr) expands to a variable name).

The reason why we don't have extern variables (and like 99.99% we don'thave extern variables) is because once they are "extern" they also become global and therefore accesible from ANY module, which would violate encapsulation.

However!! This is a "const" so all writes would fail.

v2/src/log_context_property_type_wchar_t_ptr.c
anporumb: "and everywhere" if there are other uses of malloc/other memory allocation functions

v2/devdoc/log_context_property_type_wchar_t_ptr_requirements.md
M-iceberg: done by convention in log_context.h

v2/inc/c_logging/log_context.h
M-iceberg: LOG_MAX_WCHAR_STRING_LENGTH has been used in both 2 files, move the defination to log_context_property_type_wchar_t_ptr.h line 14 with value 4096

v2/inc/c_logging/log_context.h
M-iceberg: yeah, MU_C2A works, modified to use MU_C2A

v2/inc/c_logging/log_context_property_type_wchar_t_ptr.h
M-iceberg: make sense thank you for explanation :)

v2/src/log_context_property_type_wchar_t_ptr.c
M-iceberg: modified to use static memory here, only here use malloc

v2/src/log_context_property_type_wchar_t_ptr.c
```diff
@@ -0,0 +1,148 @@
...
+
+int LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr)(const wchar_t* format, ...)
+{
+    va_list args;
+    va_start(args, format);
+
+    wchar_t buffer[LOG_MAX_WCHAR_STRING_LENGTH];
+    /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_018: [ If vswprintf fails, LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr) shall return a negative value. ]*/
+    /* Codes_SRS_LOG_CONTEXT_PROPERTY_TYPE_WCHAR_T_PTR_07_019: [ Otherwise, on success, LOG_CONTEXT_PROPERTY_TYPE_GET_INIT_DATA_SIZE(wchar_t_ptr) shall return the amount of memory needed in number of wide-chracters to store the wprintf style formatted wchar_t string given by format and the arguments in .... ]*/
+    int result = vswprintf(buffer, LOG_MAX_WCHAR_STRING_LENGTH, format, args) * sizeof(wchar_t);
```
anporumb, line 128: > vswprintf(buffer, LOG_MAX_WCHAR_STRING_LENGTH, [](http://example.com/codeflow?start=16&length=47)

I think the intention here is to only get the size. We might then call vswprintf with NULL, 0 for the first and second parameter. I suspect that the work done by `vswprintf` would be less then.

v2/src/log_context_property_type_wchar_t_ptr.c
```diff
@@ -0,0 +1,148 @@
...
+    if (result >= 0)
+    {
+        result += sizeof(wchar_t);
+    }
+    else
+    {
+        (void)printf("wchar_t string length exceeds LOG_MAX_WCHAR_STRING_LENGTH, wchar string has been truncated");
```
anporumb, line 135: >printf [](http://example.com/codeflow?start=14&length=6)

should print the number of characters.

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
```diff
@@ -0,0 +1,255 @@
...
+    int result = LOG_CONTEXT_PROPERTY_TYPE_IF_IMPL(wchar_t_ptr).copy(buffer, NULL);
+
+    // assert
+    POOR_MANS_ASSERT(result != 0);
+}
+
+static void wchar_t_ptr_copy_with_NULL_dst_value_fails(void)
+{
+    // arrange
+    wchar_t buffer[sizeof(L"cucu")];
```
anporumb, line 100: >sizeof [](http://example.com/codeflow?start=19&length=6)

sizeof(L"cucu")/sizeof(wchar_t)

anporumb: here and everywhere

anporumb (approved):
:shipit:

v2/src/log_context_property_type_wchar_t_ptr.c
M-iceberg: have deleted this printf, vswprintf doesn't state it can return number of wchars should be written by passing NULL and 0, so I use LOG_MAX_WCHAR_STRING_LENGTH. If that works, just remove this print to keep the same the ascii char.

v2/tests/log_context_property_type_wchar_t_ptr_int/log_context_property_type_wchar_t_ptr_int.c
M-iceberg: done

v2/src/log_context_property_type_wchar_t_ptr.c
M-iceberg: pass NULL and 0 doesn't work on linux, leave it as before

v2/src/log_context_property_type_wchar_t_ptr.c
M-iceberg: don't actually know the size here, vswprintf return size is limit by the second parameter

M-iceberg: vswprint returns a negative value in cases when actual length is greater than the length passed or an output error occurs, can't distinguish between them. removed this line, error will occur if result < 0 in log.context.h

PR #152, opened 2023-08-13
use GetModuleFileNameA instead of _get_pgmptr

anporumb:


anporumb:
DevOps task is 24831870

PR #156, opened 2023-09-26
Change console colors in v2 to make LOG_INFO less...yellow

mattdurak:
Changes Log Verbose to be gray, Log Info to just be default white, and Log Warning to be yellow. Seems a bit less jarring when so many LogInfo's are printed:
image


dcristoloveanu (dismissed):
:shipit:

jebrando (approved):
:shipit:

dcristoloveanu (approved):
:shipit:

PR #158, opened 2023-10-03
Add back the check for printf format (which used to be there in v1, but lost in v2)

dcristoloveanu:
Add back the check for printf format (which used to be there in v1, but lost in v2)

### DIFF
--- a/v2/inc/c_logging/logger.h
+++ b/v2/inc/c_logging/logger.h
@@ -4,6 +4,12 @@
 #ifndef LOGGER_H
 #define LOGGER_H
 
+#ifdef __cplusplus
+#include <cstdio>
+#else
+#include <stdio.h>
+#endif
+
 #include "c_logging/log_context.h"
 #include "c_logging/log_level.h"
 #include "c_logging/log_sink_if.h"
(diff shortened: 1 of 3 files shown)


v2/inc/c_logging/logger.h
```diff
@@ -50,10 +56,17 @@ extern "C" {
     void logger_log(LOG_LEVEL log_level, LOG_CONTEXT_HANDLE log_context, const char* file, const char* func, int line_no, const char* format, ...);
     void logger_log_with_config(LOGGER_CONFIG logger_config, LOG_LEVEL log_level, LOG_CONTEXT_HANDLE log_context, const char* file, const char* func, int line_no, const char* format, ...);
 
+#define LOGGER_FORMATTING_SYNTAX_CHECK(format, ...) \
+    (void)(0 && printf(format MU_IFCOMMALOGIC(MU_COUNT_ARG(__VA_ARGS__)) __VA_ARGS__));
```
anporumb, line 60: >(0 && printf(format MU_IFCOMMALOGIC(MU_COUNT_ARG(__VA_ARGS__)) __VA_ARGS__)) [](http://example.com/codeflow?start=10&length=76)

should be between do{} while(0)

anporumb, line 60: >(void) [](http://example.com/codeflow?start=4&length=6)

what happens if you remove this cast?

dcristoloveanu: I remember this was like that for a reason. It is carried over from the previous v1 logging.
But I removed it now and we'll see if it comes back. I suspect it might on some gcc flavor - but we'll deal with it then.

dcristoloveanu: Added

dcristoloveanu: Yep, as I remembered, it did fail on Linux, so adding it back.

avranju (approved):

PR #166, opened 2023-11-03
capture another thread stack

anporumb:


v2/src/log_stacktrace.c
```diff
@@ -0,0 +1,279 @@
...
+#include <string.h>
+#include <inttypes.h>
+#include <stddef.h>
+#include <stdbool.h>
+
+#include "windows.h"
+#include "winnt.h"
+#include "dbghelp.h"
+
+#include "c_logging/logging_stacktrace.h"
```
dcristoloveanu, line 14: >"c_logging/logging_stacktrace.h" [](http://example.com/codeflow?start=9&length=32)

Is this the wrong included header?

v2/src/log_stacktrace.c
```diff
@@ -0,0 +1,279 @@
...
+
+static volatile LONG doSymInit = 0; /*0 = not initialized, 1 = initializing, 2= initialized*/
```
dcristoloveanu, line 16: > volatile LONG doSymInit = 0;  [](http://example.com/codeflow?start=6&length=30)

Once the BDF suggested using an anonymous union for these - /consider

v2/src/log_stacktrace.c
```diff
@@ -0,0 +1,279 @@
...
+
+static SRWLOCK lockOverSymCalls = SRWLOCK_INIT;
+
+/*returns number of characters copied into destination*/
+static size_t memcat(char* destination, size_t destinationSize, const char* source, size_t sourceSize)
+{
+    size_t sizeToCpy = destinationSize<sourceSize?destinationSize:sourceSize;
+    (void)memcpy(destination, source, sizeToCpy);
+    return sizeToCpy;
+}
```
dcristoloveanu, line 32: Well, this is used in one place, so I could say that maybe we don't need a function and we just memcpy.

v2/src/log_stacktrace.c
```diff
@@ -0,0 +1,279 @@
...
+                        line_number = 0;
+                    }
+                    else
+                    {
+                        file_name = line.FileName;
+                        line_number = line.LineNumber;
+                    }
+
+                    /*4.3) append the function name, file name and line number to the destination*/
+                    snprintf_fallback(&destination, &destinationSize, snprintfFailed, sizeof(snprintfFailed), "%s!%s %s:%" PRIu32 "", firstLine ? (firstLine = false, "") : "\n", function_name, file_name, line_number);
```
dcristoloveanu, line 246: >snprintf_fallback [](http://example.com/codeflow?start=20&length=17)

This might be an interesting thing to extract somehow. Because it would be useful in other parts of the SW (I am thinking when building a log line - similar logic is implemented there and maybe it would not be bad to had a snprintf_with_fallback as a separate unit.

/task is fine.

v2/src/log_stacktrace.c
```diff
@@ -0,0 +1,279 @@
...
+
+                    IMAGEHLP_LINE64 line;
+                    line.SizeOfStruct = sizeof(IMAGEHLP_LINE64);
+                    DWORD lineDisplacement;
+                    const char* file_name;
+                    uint32_t line_number;
+
+                    /*4.2) determine the file name and line number*/
+                    if (!SymGetLineFromAddr64(hProcess, stackFrame.AddrPC.Offset, &lineDisplacement, &line))
+                    {
```
dcristoloveanu, line 235: Would be nice to have UTs also for this. Because likely the int tests and not injecting any kind of errors.
Likely a separate task.

v2/src/log_stacktrace.c
```diff
@@ -0,0 +1,279 @@
...
+        HANDLE hProcess = GetCurrentProcess();
+
+        /*2) call SymInitialize for the current process*/
+        /*lazily call once SymInitialize*/
+        LONG state;
+        while ((state = InterlockedCompareExchange(&doSymInit, 1, 0)) != 2)
+        {
+            if (state == 0)
+            {
+                (void)SymInitialize(hProcess, NULL, TRUE); /*this is a process-wide initialization, and will leak because we don't know when to call SymCleanup. It is a good candidate for platform_init*/
```
dcristoloveanu, line 126: > It is a good candidate for platform_init*/ [](http://example.com/codeflow?start=160&length=43)

I think this module could benefit from having a log_stacktrace_init/deinit.
It could be added to logger_init/deinit, platform_init/deinit or left to the user event (although this would likely be sucky). Likely platform init would be a good candidate.

v2/inc/c_logging/log_stacktrace.h
```diff
@@ -0,0 +1,29 @@
...
+#include "windows.h" /*needed for HANDLE*/
+#else
+#error no other platform supported
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    void getThreadStackAsString(HANDLE thread, char* destination, size_t destinationSize);
```
dcristoloveanu, line 23: >void getThreadStackAsString(HANDLE thread, char* destination, size_t destinationSize); [](http://example.com/codeflow?start=4&length=86)

Frankly this API has nothing to do with logging. It is merely linked to logging because that's where we put it the first time.

But we actually want to obtain the callstacks in some cases and save the callstack (like we do in EBS when we have a report fault).

I think it's true name is probably get_stacktrace.h/c

/poetry

dcristoloveanu (approved):
:shipit:

v2/src/log_stacktrace.c
anporumb: sheepishly:... yes

v2/src/log_stacktrace.c
anporumb: The value of doSymInit is never logged (where it makes sense to have it as a union). Nor do I expect huge debugging efforts around it.

Will leave it as is.

v2/src/log_stacktrace.c
anporumb: removed and did it inline

v2/src/log_stacktrace.c
anporumb: Changed my mind and doing it anyway.

v2/src/log_stacktrace.c
anporumb: Myeah, another task.

[https://msazure.visualstudio.com/One/_workitems/edit/25732458](https://msazure.visualstudio.com/One/_workitems/edit/25732458)

v2/src/log_stacktrace.c
anporumb: Made a task.

anporumb: [https://msazure.visualstudio.com/One/_workitems/edit/25732754](https://msazure.visualstudio.com/One/_workitems/edit/25732754)

v2/src/log_stacktrace.c
anporumb: For the current case that we have (where this "as written" works just nice) I don't see an added value to have this call to SymInitialize done in a separate fucntion.

However, we can do this with a different task, which I will put in "things that never get done 3".

anporumb: added [https://msazure.visualstudio.com/One/_workitems/edit/25733038](https://msazure.visualstudio.com/One/_workitems/edit/25733038)

dcristoloveanu (approved):

PR #182, opened 2023-12-01
Add install instructions for logging V2.

darobs:
Add a line to install libraries and headers so we can install (especially in Linux).


mattdurak (approved):
:shipit:

PR #191, opened 2024-02-26
Extract log_context_property_to_string from log_sink_console

mattdurak:
for use with other log sinks that also want to convert all of the data to text


dcristoloveanu (approved):
:shipit:

PR #192, opened 2024-02-27
Add log_sink_callback

mattdurak:
This is another log sink which encodes the logged message similar to the `log_sink_console` (without color) and passes the level and string to a callback function to handle.

This log sink allows the logger to forward messages to any other logging system (e.g. in another language like rust).

In addition to the normal `logger_init`, `log_sink_callback_set_callback` must also be called to initialize this.


v2/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@
 #the following variables are project-wide and can be used with cmake-gui
 option(log_abort_on_error "set log_abort_on_error to ON abort in case an error is encountered while using stack contexts (default is OFF)" OFF)
 option(log_sink_console "Use the log console sink (send logs to console). Default is ON" ON)
+option(log_sink_callback "Use the log callback sink (send logs to a custom callback function). Code must call log_sink_callback_set_callback. Default is OFF" OFF)
```
dcristoloveanu, line 7: >log_sink_callback [](http://example.com/codeflow?start=7&length=17)

neat-o

dcristoloveanu (approved):
:shipit:

PR #193, opened 2024-03-05
Make max log level for log_sink_callback configurable

mattdurak:
Adds function `log_sink_callback_set_max_level` to set the max log level.
No string processing is done when log level is above the max.


darobs (approved):
:shipit:

PR #229, opened 2024-11-05
Add stdarg include here because I ran into a compiler error.

darobs:
CLANG complained that va_list was used but not defined.


dcristoloveanu (approved):
:shipit:

PR #248, opened 2025-07-16
[Copilot] Remove v1 logging artefacts

dcristoloveanu:
Remove v1 logging artefacts


inc/c_logging/consolelogger.h
```diff
@@ -1,25 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved.
```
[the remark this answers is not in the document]
dcristoloveanu: We'll move it, just did not want to do it all at once

v2/src/logging_stacktrace.c
```diff
@@ -0,0 +1,45 @@
...
+
+/*tries to get as much as possible from the stack filling destination*/
+void getStackAsString(char* destination, size_t destinationSize)
+{
+    /*Codes_SRS_LOGGING_STACKTRACE_01_002: [ If destination is NULL, getStackAsString shall return without doing anything. ]*/
+    /*Codes_SRS_LOGGING_STACKTRACE_01_003: [ If destinationSize is 0, getStackAsString shall return without doing anything. ]*/
+    if (destination == NULL || destinationSize == 0)
+    {
+        return;
+    }
```
jebrando, line 28: Interesting that you are returning in the middle of a function.  Would it be better to just have an else after this and then return at the bottom?

jebrando (approved):
:shipit:

v2/src/logging_stacktrace.c
dcristoloveanu: Changed

PR #316, opened 2026-04-01
get_thread_stack should be part of logger_init

anporumb:


parth21999 (dismissed):
:shipit:

v2/inc/c_logging/get_thread_stack.h
```diff
@@ -20,7 +22,17 @@
 extern "C" {
 #endif
 
+    /*initializes statics needed to get a thread stack - not thread safe in itself*/
+    int get_thread_stack_init(void);
+
+#if defined(_MSC_VER)
     void get_thread_stack(DWORD threadId, char* destination, size_t destinationSize);
```
dcristoloveanu: >DWORD [](http://example.com/codeflow?start=26&length=5)

I am thinking that If we would have a C_LOGGING_THREAD_T or similar then we would have one signature, since the only different thing at signature level is the type for the thread.

v2/src/get_thread_stack.c
```diff
@@ -1,54 +1,36 @@
...
-static SYM_INIT_STATE doSymInit = { .state_enum = SYM_INIT_NOT_INITIALIZED };
-
-typedef union SYMBOL_INFO_EXTENDED_TAG
-{
-    SYMBOL_INFO symbol;
-    unsigned char extendsUnion[sizeof(SYMBOL_INFO) + MAX_SYM_NAME - 1]; /*this field only exists to extend the size of the union to encompass "CHAR    Name[1];" of the SYMBOL_INFO to be as big as TRACE_MAX_SYMBOL_SIZE - 1 */ /*and the reason why it is not malloc'd is to exactly avoid a malloc that cannot be LogError'd (how does one log errors in a log function?!)*/
-}SYMBOL_INFO_EXTENDED;
-
-static SRWLOCK lockOverSymCalls = SRWLOCK_INIT;
+/*note: the APIs in here cannot use LogError+friends because it would result in a stack overflow. Here's why.
```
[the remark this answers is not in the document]
dcristoloveanu: So probably we should have UTs (or a task to implement them somewhere)

dcristoloveanu (dismissed):
:shipit:

v2/inc/c_logging/get_thread_stack.h
anporumb: If I change this... to be C_LOGGING_THREAD_T then every Win API will take this weird type (even when typedef'd from DWORD).

I think we will know better about these types when we will actualyl have the same functionality on Linux. 

:)

So let's wait until we have a linux impl here, see how that goes.

dcristoloveanu (approved):
:shipit:

PR #329, opened 2026-04-09
[MrBot] Add get_thread_stack_refresh_module_list for post-init symbol refresh

mattdurak:
## Problem

When `get_thread_stack_init()` / `SymInitialize()` runs before test DLLs are loaded (e.g. in VsTest testhost processes), `SymFromAddr()` cannot resolve function names from those DLLs. This causes `wait_for_callstack` in EBS to return `"failure in SymFromAddr"` instead of real function names, leading to spurious `WAIT_FOR_CALLSTACK_TIMEOUT` failures.

## Fix

Add `get_thread_stack_refresh_module_list()` which calls `SymRefreshModuleList()` to update dbghelp's loaded module list. This is a targeted, low-risk addition — no existing behavior changes.

Consumers (like EBS's `wait_for_callstack`) call this once before their polling loop to ensure recently-loaded DLLs have symbols available.

## Testing

Will be validated via EBS gate PR that points c-logging submodule to this branch.

### DIFF
--- a/v2/src/get_thread_stack.c
+++ b/v2/src/get_thread_stack.c
@@ -320,6 +320,7 @@ void get_thread_stack_refresh_module_list(void)
                                     {
                                         searchPath[searchPathLen++] = ';';
                                     }
+
                                     if (searchPathLen + dirLen < sizeof(searchPath))
                                     {
                                         (void)memcpy(searchPath + searchPathLen, dir, dirLen);
--- a/v2/src/get_thread_stack.c
+++ b/v2/src/get_thread_stack.c
@@ -253,12 +253,29 @@ VsTest into testhost.exe) have their PDBs in different directories. This functio
 3. re-initializes the symbol handler with the updated search path so SymFromAddr can resolve their functions */
 void get_thread_stack_refresh_module_list(void)
 {
-    if (g.symbolsState == SYM_INIT_INITIALIZED)
+    if (g.symbolsState != SYM_INIT_INITIALIZED)
+    {
+        /*not initialized, nothing to refresh*/
+    }
+    else
     {
         AcquireSRWLockExclusive(&g.lockOverSymCalls);
         {
             HANDLE hModuleSnap = CreateToolhelp32Snapshot(TH32CS_SNAPMODULE, 0);
-            if (hModuleSnap != INVALID_HANDLE_VALUE)
+            if (hModuleSnap == INVALID_HANDLE_VALUE)
+            {
+                /*fall back to just refreshing the module list*/
+                if (!SymRefreshModuleList(g.processHandle))
+                {
+                    (void)printf("failure (GetLastError()=0x%" PRIx32 ") in SymRefreshModuleList(g.processHandle=%p)\n",
+                        GetLastError(), g.processHandle);
+                }
+                else
+                {
+                    /*SymRefreshModuleList succeeded*/
+                }
+            }
+            else
             {
                 char searchPath[32768];
                 size_t searchPathLen = 0;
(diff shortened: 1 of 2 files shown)


v2/src/get_thread_stack.c
```diff
@@ -244,6 +245,88 @@ int get_thread_stack_init(void)
...
+                if (searchPathLen > 0)
+                {
+                    /*re-initialize the symbol handler with the updated search path*/
+                    (void)SymCleanup(g.processHandle);
+                    if (!SymInitialize(g.processHandle, searchPath, TRUE))
+                    {
+                        (void)printf("failure (GetLastError()=0x%" PRIx32 ") in SymInitialize during refresh(g.processHandle=%p, searchPath=%s, TRUE)\n",
+                            GetLastError(), g.processHandle, searchPath);
+                    }
+                }
```
mattdurak: >} [](http://example.com/codeflow?start=16&length=1)

always have an "else" even if it's just a comment

v2/src/get_thread_stack.c
```diff
@@ -244,6 +245,124 @@ int get_thread_stack_init(void)
...
+                                if (strstr(searchPath, dir) != NULL)
+                                {
+                                    /*directory already in search path, skip*/
+                                }
+                                else
+                                {
+                                    if (searchPathLen > 0 && searchPathLen + 1 < sizeof(searchPath))
+                                    {
+                                        searchPath[searchPathLen++] = ';';
+                                    }
```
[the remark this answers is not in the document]
mattdurak: 🤖 MrBot: Fixed in fd32821 — added blank line between the two if blocks.

v2/src/get_thread_stack.c
```diff
@@ -244,6 +245,125 @@ int get_thread_stack_init(void)
...
+
+                if (searchPathLen == 0)
+                {
+                    /*no directories found, nothing to update*/
+                }
+                else
+                {
+                    /*re-initialize the symbol handler with the updated search path*/
+                    (void)SymCleanup(g.processHandle);
+                    if (!SymInitialize(g.processHandle, searchPath, TRUE))
```
anporumb, line 350: >SymInitialize [](http://example.com/codeflow?start=25&length=13)

So... you're not going to like this , but I believe that the right way to do this is not be cleanup+initialize but SymSetSearchPath should help better here (and everywhere here).

SymInitialize with search path NULL has value because it then looks in env variables for symbols. It is unclear to me if it still does that once the searchPath is set.

anporumb: If I am wrong, just check in, I will sign off

anporumb (approved):
:shipit:

v2/src/get_thread_stack.c
mattdurak: 🤖 MrBot: Valid point — SymInitialize with NULL preserves default search behavior (env variables + module directory), and SymSetSearchPath is less heavy than SymCleanup+SymInitialize. I'll create an experiment branch using SymSetSearchPath instead, point the EBS PR to it, and run the gate a few times to validate.

mattdurak: 🤖 MrBot: Tested the SymSetSearchPath approach — **it failed 3/3 runs** ([160015688](https://msazure.visualstudio.com/One/_build/results?buildId=160015688), [160015690](https://msazure.visualstudio.com/One/_build/results?buildId=160015690), [160015691](https://msazure.visualstudio.com/One/_build/results?buildId=160015691)).

The issue: \SymSetSearchPath\ + \SymRefreshModuleList\ updates the search path and finds new modules, but does **not re-load PDBs** for modules that were already enumerated with the stale path during \SymInitialize\. The module entries are already in dbghelp's internal list — \SymRefreshModuleList\ only adds/removes modules, it doesn't retry PDB loading for existing ones.

The \SymCleanup\ + \SymInitialize\ approach is necessary because it fully re-initializes the symbol handler, forcing fresh PDB lookups for all modules with the updated search path. This approach went **4/4 green** (builds [159911616](https://msazure.visualstudio.com/One/_build/results?buildId=159911616), [159911637](https://msazure.visualstudio.com/One/_build/results?buildId=159911637), [159918534](https://msazure.visualstudio.com/One/_build/results?buildId=159918534), [159918538](https://msazure.visualstudio.com/One/_build/results?buildId=159918538)).

Reverting EBS back to the SymCleanup+SymInitialize PR branch.

mattdurak: For reference, the SymSetSearchPath experiment branch is [users/mdurak/sym-set-search-path-experiment](https://github.com/Azure/c-logging/tree/users/mdurak/sym-set-search-path-experiment) (commit 03e6722).

PR #338, opened 2026-04-29
Add macOS (APPLE) stub support for get_thread_stack

ewertons:
macOS uses pthreads like Linux. The stub implementation returns 0 for init and prints 'not supported' for stack traces, matching the existing Linux behavior. This unblocks c-utility and downstream repos building and running tests on macOS.


dcristoloveanu (approved):
:shipit:

PR #375, opened 2026-07-28
Fix unbounded %s reused as sscanf format in log sink test helpers

anporumb:
validate_log_line and validate_log_line_with_NULL_time build an expected_string via snprintf using the caller supplied literal format (which contains escaped %%s/%%d tokens), then reuse that string directly as the sscanf format to parse the actual log output. The surviving %s (and %[^\r\n]) conversions had no field width, so they could write past the small stack buffers used to receive them (day_of_week[4], month[4], reset_color_code[10], anything_else[200], message[LOG_MAX_MESSAGE_LENGTH]) if the actual output ever exceeded those sizes.

Added explicit field widths (matching each destination buffer size minus 1 for the NUL terminator) to every call site literal in log_sink_console_ut.c and log_sink_callback_ut.c.

Verified by building and running both log_sink_console_ut.exe and log_sink_callback_ut.exe (exit code 0, all assertions pass).

### DIFF
--- a/v2/tests/log_sink_callback_ut/log_sink_callback_ut.c
+++ b/v2/tests/log_sink_callback_ut/log_sink_callback_ut.c
@@ -626,7 +626,7 @@ static void log_sink_callback_log_calls_callback_with_one_CRITICAL_log_line(void
     POOR_MANS_ASSERT(actual_and_expected_match);
     POOR_MANS_ASSERT(expected_calls[actual_call_count - 1].log_callback_call.captured_context == (void*)0x42);
     POOR_MANS_ASSERT(expected_calls[actual_call_count - 1].log_callback_call.captured_log_level == LOG_LEVEL_CRITICAL);
-    validate_log_line(expected_calls[actual_call_count - 1].log_callback_call.captured_output, "Time:%%s %%s %%d %%d:%%d:%%d %%d File:%s:%d Func:%s %%[^\r\n]", __FILE__, line_no, __FUNCTION__, "test");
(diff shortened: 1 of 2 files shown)


v2/tests/log_sink_callback_ut/log_sink_callback_ut.c
```diff
@@ -626,7 +626,7 @@ static void log_sink_callback_log_calls_callback_with_one_CRITICAL_log_line(void
     POOR_MANS_ASSERT(actual_and_expected_match);
     POOR_MANS_ASSERT(expected_calls[actual_call_count - 1].log_callback_call.captured_context == (void*)0x42);
     POOR_MANS_ASSERT(expected_calls[actual_call_count - 1].log_callback_call.captured_log_level == LOG_LEVEL_CRITICAL);
-    validate_log_line(expected_calls[actual_call_count - 1].log_callback_call.captured_output, "Time:%%s %%s %%d %%d:%%d:%%d %%d File:%s:%d Func:%s %%[^\r\n]", __FILE__, line_no, __FUNCTION__, "test");
+    validate_log_line(expected_calls[actual_call_count - 1].log_callback_call.captured_output, "Time:%%3s %%3s %%d %%d:%%d:%%d %%d File:%s:%d Func:%s %%4095[^\r\n]", __FILE__, line_no, __FUNCTION__, "test");
```
mattdurak, line 629: >4095 [](http://example.com/codeflow?start=152&length=4)

Could we pass this in as a define so it's a bit less of a magic number? Here and everywhere

mattdurak (approved):
:shipit:

